PR #72, opened 2023-05-03
Add features for python-lit and ICC.

wmaxey:
This adds some pretty simple features for installing lit and icc.

Lit is simple, but I had to force icc onto the path. nvcc+icc don't seem to work with sccache.


features/src/python-lit/install.sh
```diff
@@ -0,0 +1,43 @@
...
+    apt-utils                   \
+    gettext-base                \
+    bash-completion             \
+    gcc g++ gfortran            \
+    software-properties-common  \
+    ;
+
+apt-get install -y --no-install-recommends python3 python3-dev python3-wheel python3-pip;
+python3 -m pip install wheel;
+python3 -m pip install psutil $LIT_VERSION_TO_INSTALL;
```
trxcllnt, line 31: Features can't currently depend on other features, but I think it's fine to assume python is already installed (e.g. via our `mambaforge` feature, `ghcr.io/devcontainers/features/python`, etc.). I do something similar in the [`rapids-build-utils`](https://github.com/rapidsai/devcontainers/blob/f854296b98ae9feae8147efd5bc2cc91c6361503/features/src/rapids-build-utils/install.sh#L25-L28) feature.

```suggestion
if ! type python >/dev/null 2>&1; then
  echo "lit feature expects python to already be installed" >&2;
  exit 1;
fi
python -m pip install $LIT_VERSION_TO_INSTALL;
```

features/src/icc/.bashrc
```diff
@@ -0,0 +1,2 @@
+export ICC_VERSION="${ICC_VERSION}";
+export PATH="$PATH";
```
trxcllnt, line 2: I've started doing this pattern to avoid modifying PATH if it already has what we want:

```suggestion
if [ -n "${PATH##*"/opt/intel/oneapi/compiler/${ICC_VERSION}/linux/bin"*}" ]; then
    export PATH="$PATH:/opt/intel/oneapi/compiler/${ICC_VERSION}/linux/bin";
fi
```

features/src/icc/install.sh
```diff
@@ -0,0 +1,37 @@
...
+wget -O- https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB | gpg --dearmor | tee /usr/share/keyrings/oneapi-archive-keyring.gpg > /dev/null;
+echo "deb [signed-by=/usr/share/keyrings/oneapi-archive-keyring.gpg] https://apt.repos.intel.com/oneapi all main" | tee /etc/apt/sources.list.d/oneAPI.list;
+apt-get update; apt-get install -y --no-install-recommends intel-dpcpp-cpp-compiler-${ICC_VERSION};
+
+export PATH="$PATH:/opt/intel/oneapi/compiler/${ICC_VERSION}/linux/bin";
+export ICC_VERSION="${ICC_VERSION}";
+
+# export envvars in bashrc files
+append_to_etc_bashrc "$(cat .bashrc | envsubst)";
+append_to_all_bashrcs "$(cat .bashrc | envsubst)";
```
trxcllnt, line 31: Only substitute references to `$ICC_VERSION`:

```suggestion
append_to_etc_bashrc "$(cat .bashrc | envsubst '$ICC_VERSION')";
append_to_all_bashrcs "$(cat .bashrc | envsubst '$ICC_VERSION')";
```

features/src/python-lit/install.sh
```diff
@@ -0,0 +1,43 @@
...
+
+export LIT_VERSION="$(lit --version | grep -o -e '[0-9].*')";
+
+# export envvars in bashrc files
+append_to_etc_bashrc "$(cat .bashrc | envsubst)";
+append_to_all_bashrcs "$(cat .bashrc | envsubst)";
```
trxcllnt, line 37: New functionality from #69:

```suggestion
append_to_all_bashrcs "$(cat .bashrc | envsubst)";
# export envvars in /etc/profile.d
add_etc_profile_d_script python-lit "$(cat .bashrc | envsubst)";

```

features/src/python-lit/install.sh
```diff
@@ -26,7 +26,10 @@ check_packages                  \
     software-properties-common  \
     ;
 
-apt-get install -y --no-install-recommends python3 python3-dev python3-wheel python3-pip;
+if ! type python >/dev/null 2>&1; then
+  echo "lit feature expects python to already be installed" >&2;
+  exit 1;
+fi
 python3 -m pip install wheel;
```
wmaxey, line 33: pip errors from some kind of dependency thing if you try to install `wheel` and `lit` at the same time. This is only necessary on 18.04 and maybe 20.04?

`psutil` is needed for `lit` to kill tests that are timing out.

features/src/python-lit/install.sh
wmaxey: Yeah, I felt weird calling this `python-lit` because I'd assume we won't use `mambaforge` in our matrix.

wmaxey:
Should I add entries for these tools in the image matrix as well?

features/src/icc/devcontainer-feature.json
```diff
@@ -0,0 +1,19 @@
...
+  "options": {
+    "version": {
+      "type": "string",
+      "proposals": [
+        "2022.2.0",
+        "2022.2.1",
+        "2023.0.0",
+        "2023.1.0"
+      ],
+      "default": "2022.2.1",
```
trxcllnt, line 15: Any reason the default is 2022 and not 2023?

wmaxey: CTK doesn't yet support 2023 if I recall correctly.

features/src/icc/install.sh
```diff
@@ -0,0 +1,39 @@
+#! /usr/bin/env bash
+set -ex
+
+ICC_VERSION="${VERSION:-$DEFAULT}";
```
trxcllnt, line 4: what is `$DEFAULT` here?

wmaxey: So this was my guess:

It's the default listed in the feature json. I thought it would get defined in the env for the script.

wmaxey:
```
#14 30.32 Processing triggers for dbus (1.12.20-2ubuntu4.1) ...
#14 30.36 + type python
#14 30.36 + echo 'lit feature expects python to already be installed'
#14 30.36 lit feature expects python to already be installed
#14 30.36 + exit 1
#14 30.36 ERROR: Feature "./python-lit" (Unknown) failed to install!
```

Looks like the test fails because Python wasn't installed.

wmaxey:
```
[2023-05-03T22:02:14.650Z] #14 18.59 Get:5 https://apt.repos.intel.com/oneapi all InRelease [4447 B]
#14 18.69 Err:5 https://apt.repos.intel.com/oneapi all InRelease
#14 18.69   The following signatures couldn't be verified because the public key is not available: NO_PUBKEY ACFA9FC57E6C5DBE

[2023-05-03T22:02:14.950Z] #14 18.91 Reading package lists...
[2023-05-03T22:02:15.701Z] 
#14 19.66 W: GPG error: https://apt.repos.intel.com/oneapi all InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY ACFA9FC57E6C5DBE
#14 19.66 E: The repository 'https://apt.repos.intel.com/oneapi all InRelease' is not signed.
#14 19.66 ERROR: Feature "./icc" (Unknown) failed to install!
```

I'm not quite sure the cause of this one.

features/src/python-lit/install.sh
```diff
@@ -0,0 +1,48 @@
...
+    gcc g++ gfortran            \
+    software-properties-common  \
+    ;
+
+if ! type python >/dev/null 2>&1; then
+  echo "lit feature expects python to already be installed" >&2;
+  exit 1;
+fi
+python3 -m pip install wheel;
+python3 -m pip install psutil $LIT_VERSION_TO_INSTALL;
```
trxcllnt, line 34:
```suggestion
python -m pip install wheel;
python -m pip install psutil $LIT_VERSION_TO_INSTALL;
```

features/src/python-lit/install.sh
trxcllnt: > pip errors from some kind of dependency thing if you try to install `wheel` and `lit` at the same time

Curious why we need `wheel`? Is lit compiled from an sdist, or does it implicitly depend on `wheel` but omit it from its build requirements?

features/src/python-lit/install.sh
```diff
@@ -0,0 +1,48 @@
...
+
+check_packages                  \
+    gpg                         \
+    dirmngr                     \
+    apt-utils                   \
+    gettext-base                \
+    bash-completion             \
+    gcc g++ gfortran            \
+    software-properties-common  \
+    ;
```
trxcllnt, line 27: Does `lit` need all these dependencies, or are these needed in order to build/install lit/wheel/psutil etc.?

PR #83, opened 2023-06-22
Add gcc6 feature and container

jrhemstad:
We need a gcc6 container for CCCL CI.

### DIFF
--- a/matrix.yml
+++ b/matrix.yml
@@ -46,6 +46,7 @@ include:
 - os: "ubuntu:20.04"
   images:
   - features: [*oneapi_2022, *cuda_curr_max, *python, *lit]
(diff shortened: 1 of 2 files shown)


matrix.yml
```diff
@@ -44,6 +46,7 @@ include:
 - os: "ubuntu:20.04"
   images:
   - features: [*oneapi_2022, *cuda_curr_max, *python, *lit]
+  - features: [*gcc_6, *cuda_curr_max, *python, *lit]
```
miscco, line 49: gcc-6 has no installable packages for 20.04
```suggestion
```

jrhemstad: Switched to making the gcc6 + cuda12.1 container based on 18.04.

PR #127, opened 2023-08-28
fix pip solve

cwharris:
Fixes the pip solve by excluding _all_ python libraries declared in the manifest rather just the ones with an associated `depedencies.yaml` file (such as the ones found in the RAPIDS repositories).


trxcllnt (approved):

features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-pip-env.sh
```diff
@@ -75,6 +64,17 @@ make_pip_env() {
         fi
     done
 
+    # add all pytohn packages to the noinstall list.
```
trxcllnt, line 67:
```suggestion
    # add all python packages to the noinstall list.
```

PR #144, opened 2023-09-19
Add options to install cuDNN, cuTensor, NCCL, and dev packages

trxcllnt:


features/src/cuda/devcontainer-feature.json
```diff
@@ -15,11 +16,32 @@
...
         "11.3",
-        "11.2"
+        "11.2",
+        "11.1"
       ],
-      "default": "12.1",
+      "default": "12.2",
       "description": "Version of the CUDA Toolkit to install."
     },
+    "installcuDNN": {
```
jrhemstad, line 25: What about another option for `installMathLibs` for things like cublas, cufft, cusolver, cusparse? Or will those implicitly be excluded if we don't install cuDNN/cuTensor?

PR #159, opened 2023-10-09
Only remove newly installed llvm packages and not always everything

miscco:
We want to install a unique version of clang-format for our devcontainer

However, the current install script will remove all prior installs of llvm tools of a different version.

Rather than nuking everything we only replace the package we installed


miscco:
After some tinkering I just try to not remove clang-format alternatives

PR #166, opened 2023-10-25
Use env.RUNNER_ENVIRONMENT to check runner type

jjacobelli:
Instead of checking if the runner name contains `linux-`, we now check the value of `env.RUNNER_ENVIRONMENT`. This value should be `github-hosted` for GitHub runners and `self-hosted` for RAPIDS runners.

Test pipeline: https://github.com/rapidsai/devcontainers/actions/runs/6639045230?pr=166

### DIFF
--- a/.github/actions/devcontainer-json/action.yml
+++ b/.github/actions/devcontainer-json/action.yml
@@ -21,8 +21,8 @@ runs:
   using: composite
   steps:
 
-    - name: Setup Node.js
-      if: contains(runner.name, 'linux-') == true
+    - if: env.RUNNER_ENVIRONMENT != 'github-hosted'
+      name: Setup Node.js
       uses: actions/setup-node@v3
       with:
         node-version: '16'
--- a/.github/actions/setup-runner-env/action.yml
+++ b/.github/actions/setup-runner-env/action.yml
@@ -6,26 +6,26 @@ runs:
   using: composite
   steps:
 
-    - if: contains(runner.name, 'linux-') != true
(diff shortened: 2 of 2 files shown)


ajschmidt8 (approved):

trxcllnt:
@jjacobelli it [doesn't look like](https://github.com/trxcllnt/devcontainers/pull/10) github-hosted runners set the `env.RUNNER_ENVIRONMENT` variable.

Is this something we only do in our self-hosted runners, or did something about the github-hosted runners change? I thought the env context propagates to actions, but I could also be mistaken.

jjacobelli:
Hey @trxcllnt, I just checked a workflow running on our runners and the variable is still defined:
image
That's not something we manually set on our side so maybe GH changed something on their runners?

trxcllnt:
Ok, I'll change the checks to `if: ${{ env.RUNNER_ENVIRONMENT == 'self-hosted' }}`

trxcllnt:
[Doesn't seem like](https://github.com/rapidsai/devcontainers/actions/runs/7780625985/job/21213662507#step:5:27) `env.RUNNER_ENVIRONMENT` is defined in our runners. I even made sure to [pass it in from the calling workflow](https://github.com/rapidsai/devcontainers/pull/221/commits/09b3e0327883605b5de38f0c9d2b6ee9168e54b5#diff-f43ab833543dfad5165700790c922c6c8d279724e93ef56089508ddcdbb324e8R38), just in case something changed like actions don't inherit the workflow environment. What am I doing wrong?

PR #167, opened 2023-10-25
Add CUDAHOSTCXX envvar to gcc-env/clang-env

jrhemstad:
In CCCL containers we need `CUDAHOSTCXX` to be populated to point to whatever the host compiler is. This envvar controls the cmake host compiler for CUDA builds. 

The easiest way to do this was to just update the `gcc-env` and `clang-env`. I purposefully avoided setting this for the icc and nvhpc images for now because we aren't using those yet.


cwharris:
LGTM

cwharris (approved):

miscco (approved):

trxcllnt (approved):

PR #173, opened 2023-11-10
Add CUDA 11.8 devcontainer to CCCL matrix

sleeepyjack:
This PR adds a single image to the CCCL devcontainer matrix which is needed to run cuCollections CI tests.

CC @jrhemstad @bdice

### DIFF
--- a/matrix.yml
+++ b/matrix.yml
@@ -77,6 +77,7 @@ include:
   images:
   - {features: [*nvhpc_prev, *python, *cccl_dev], env: *nvhpc_env}
   - {features: [*nvhpc_curr, *python, *cccl_dev], env: *nvhpc_env}
+  - {features: [*gcc_11, {<<: *cuda_prev_max, <<: *cccl_cuda_opts}, *python, *cccl_dev], env: *gcc_env}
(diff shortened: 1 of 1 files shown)


matrix.yml
```diff
@@ -77,6 +77,7 @@ include:
   images:
   - {features: [*nvhpc_prev, *python, *cccl_dev], env: *nvhpc_env}
   - {features: [*nvhpc_curr, *python, *cccl_dev], env: *nvhpc_env}
+  - {features: [*gcc_11, {<<: *cuda_prev_max, <<: *cccl_cuda_opts}, *python, *cccl_dev], env: *gcc_env}
```
miscco, line 80: You put them into the CCCL bracket. Would it make sense to add a separate block with cuCollections as a target.

With that container in here, we are running risk that we might change the devcontainer when we do changes to the CCCL ones

sleeepyjack: For cuCollections we try to adhere to the CCCL setup as close as possible. Although, we only test a small subset of the CCCL matrix. I'm fine with leaving this in the CCCL section, since all of the other images in our matrix are CCCL images anyway.

miscco (approved):

miscco:
I started a deployment so that the image gets pushed to the registry

https://github.com/rapidsai/devcontainers/actions/runs/6827022148

PR #182, opened 2023-12-04
Hack in an updated GCC13 container.

wmaxey:


features/src/gcc/install.sh
```diff
@@ -25,6 +25,7 @@ GCC_VERSION="${VERSION:-${GCC_VERSION_DEFAULT}}";
 
 if [[ "${GCC_VERSION}" != "${GCC_VERSION_DEFAULT}" ]]; then
     apt-add-repository -y ppa:ubuntu-toolchain-r/test;
+    apt-add-repository -y ppa:ubuntu-toolchain-r/ppa;
```
wmaxey, line 28: This enables installing GCC 13.2 as opposed to GCC 13.1 in the test PPA.

wmaxey:
I rebased to be up to date with the target branch. I think that the build failure last seen were not related my to change, and seemed more like timeouts with various repos.

jrhemstad:
I figured out why the gcc13 jobs are failing.

I figured out why your gcc13 changes are failing.

It's ultimately because there is no Ubuntu 23.04 CTK package here: https://developer.download.nvidia.com/compute/cuda/repos/

So that breaks this: https://github.com/rapidsai/devcontainers/blob/99da203868d529cae7c9364f55cdd15aa77ba052/features/src/cuda/install.sh#L25-L52

features/src/cuda/install.sh
```diff
@@ -36,6 +36,10 @@ get_cuda_deb() {
     | grep -P "^Filename: \./${2}(.*)\.deb$"      \
     | sort -Vr | head -n1 | cut -d' ' -f2         \
     )";
+    if [ -z "$deb" ]; then
+        echo "Error: No matching .deb found for '${1}' and '${2}'" >&2
+        return 1
+    fi
```
trxcllnt, line 42: This won't work for Ubuntu 23.04, we'll need some way to use the CUDA Ubuntu 22.04 apt repo.

Issue #183, opened 2023-12-04
Add feature for Clang for NVIDIA Grace CPU


jrhemstad:
I would like a devcontainer feature to install the Grace clang compiler: https://developer.nvidia.com/grace/clang/downloads

I would also like a CCCL image added for using Grace clang as a host compiler.

wmaxey:
I think the only difficulty with the Grace compiler is that it's ARM64 exclusive. OneAPI was a pain because of that.

PR #184, opened 2023-12-05
Add clang grco for CCCL

wmaxey:
Addresses #183


features/src/llvm/install.sh
```diff
@@ -22,26 +22,55 @@ check_packages                  \
...
-    LLVM_VERSION="$(echo $LLVM_VERSION | grep -oP '[0-9]+')";
-fi
+install_clang_grace() {
+    if [[ "$(uname -p)" != "aarch64" ]]; then
+        echo "Clang for Grace is only supported on aarch64"
+        exit 0;
+    fi
+
+    echo "Installing NVIDIA Clang for Grace"
+    wget -O clang-grace-toolchain.deb https://developer.nvidia.com/downloads/assets/grace/clang/16.23.08/clang-grace-toolchain_16.23.08_arm64.deb
```
jrhemstad, line 32: Question: Are we okay hard coding the version for now?

wmaxey: I didn't see another version publicly. We can devise a mechanism later probably.

jrhemstad (approved):

jrhemstad (changes requested):
@trxcllnt requested that we make clang.grco a separate feature instead of part of the existing llvm feature.

PR #191, opened 2023-12-16
Add wholegraph to `manifest.yaml`

trxcllnt:
Adds [rapidsai/wholegraph](https://github.com/rapidsai/wholegraph) to `manifest.yaml`.

`wholegraph` depends on:
* `clang-tools=16.0.0` instead of `clang-tools=16.0.6`
* `doxygen=1.8.20` instead of `doxygen=1.9.1`
* `cudnn=8.4` instead of `cudnn=8.8`

Updating those causes the combined env to solve, not sure yet whether it builds.

### DIFF
--- a/.devcontainer/build-rapids.sh
+++ b/.devcontainer/build-rapids.sh
@@ -26,72 +26,71 @@ build_rapids() {
     (
         echo "building RMM";
         clean-rmm;
-        build-rmm -DBUILD_BENCHMARKS=ON;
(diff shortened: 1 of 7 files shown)


bdice (approved):
This PR helped fix the problems with pulling in `libraft` in the conda environment. I merged a PR to wholegraph that fixes the version inconsistencies, so this should be ready to merge.

PR #206, opened 2024-01-10
Fully set up ucxx

vyasr:
This PR adds some missing pieces from #203 to get ucxx working properly in devcontainers. Note that this PR is not sufficient for builds to succeed, there look to be some additional changes required in ucxx itself.

CC @msarahan

### DIFF
--- a/.devcontainer/build-rapids.sh
+++ b/.devcontainer/build-rapids.sh
@@ -38,6 +38,13 @@ build_rapids() {
     ) 2>&1 | maybe_write_build_log kvikio;
 
     (
+        echo "building ucxx";
+        clean-ucxx;
+        build-ucxx --verbose;
+        sccache -s;
+    ) 2>&1 | maybe_write_build_log ucxx;
+
+    (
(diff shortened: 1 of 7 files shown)


vyasr:
There's one other minor wrinkle here that I just noticed. We now get some warnings from https://github.com/rapidsai/devcontainers/blob/branch-24.02/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-vscode-workspace.sh#L14 because adding ucxx means we now have 10 repos, and `sort` and `join` don't agree on what constitutes an ordered set with the current options. The join docs recommend using `sort -k 1b,1`, but that doesn't quite seem sufficient (it only removes 1/2 warnings). I'll report back when I figure out why.

msarahan:
Linking slack thread for context: https://nvidia.slack.com/archives/C01PF84NXCL/p1704853676621399?thread_ts=1704708016.899789&cid=C01PF84NXCL

msarahan (approved):

PR #228, opened 2024-02-08
Update to sccache v0.7.7

trxcllnt:
Update sccache to fix https://github.com/mozilla/sccache/issues/2038


bdice (commented):
Looking at the last update: https://github.com/rapidsai/devcontainers/commit/aba251dc6e9c4f7679752f5bf3bee916792d825c.

Do we also need to update `features/src/sccache/devcontainer-feature.json` to remove `0.7.4` as a candidate feature version, and replace it with `0.7.7`?

KyleFromNVIDIA:
IMO, there should be no trace of `0.7.4` or `0.7.5` being allowed to be used *anywhere*. We should replace it.

features/src/sccache/devcontainer-feature.json
bdice, line 1: Does this feature need to have its version bumped?

bdice (approved):

features/src/sccache/devcontainer-feature.json
trxcllnt: Yes it does.

PR #234, opened 2024-02-09
use self-hosted windows runners

trxcllnt:
@wmaxey Are you still testing multiple versions of windows in CCCL CI? If not, can we drop the Windows Server 2019 builds and prune anything from the [Windows matrix](https://github.com/rapidsai/devcontainers/blob/branch-24.04/matrix.yml#L96-L114)?


wmaxey:
Windows 2019 is still needed for people running on Windows 10. We don't use it in CI.

trxcllnt:
@wmaxey do you think we could add `gitbash.exe` to the [Windows image](https://github.com/nv-gha-runners/vm-images/pull/33)? The GitHub hosted runners [have it](https://github.com/actions/runner-images/blob/win22/20240211.1/images/windows/Windows2022-Readme.md#shells), and it looks like that's what's blocking using the runners in this PR.

I'd make a PR, but I don't know anything about Windows.

wmaxey:
We can build 2019 on 2022 images. Builds will happen in Hyper-V so there isn't any issue with compatibility there.

PR #236, opened 2024-02-14
Requesting a clean cpp build also clears any Jitify cache

robertmaynard:
This is needed for libcudf specifically, but written to be generalized in case other projects start using Jitify as well.

### DIFF
--- a/features/src/rapids-build-utils/opt/rapids-build-utils/bin/tmpl/cpp-clean.tmpl.sh
+++ b/features/src/rapids-build-utils/opt/rapids-build-utils/bin/tmpl/cpp-clean.tmpl.sh
@@ -5,6 +5,9 @@ clean_${CPP_LIB}_cpp() {
     set -euo pipefail;
 
     rm -rf ~/${CPP_SRC}/build/latest/*;
+    if test -f "~/.${CPP_LIB}/"; then
+        rm -rf "~/.${CPP_LIB}/"
+    fi
 }
 
 if test -n "${rapids_build_utils_debug:-}"; then


trxcllnt:
Yeah, I noticed this while building DLFW and [fixed it](https://github.com/rapidsai/devcontainers/pull/221/files#diff-1eacc0167366d4e08cb7ef59b7f596f5cf1b8d8650df5661d102a960c01ed7bdR27-R30) over in https://github.com/rapidsai/devcontainers/pull/221 by deleting `<build-dir>/{*,.*}`.

We should probably do the same thing here.

robertmaynard:
is dev containers changing the local cache location? otherwise the PR DLFW looks wrong as the cudf cache (by default) is in `$HOME/.cudf/<version>`

trxcllnt:
@robertmaynard Ah, I missed that. No, devcontainers does not change the cache location.

It seems risky for the `clean-<repo>-cpp` command to delete things outside the build dir.

Can we change the cache location to under the build dir (it's controllable via envvar, right?), so then it gets deleted along with the builddir contents. #221 adds the ability to define envvars for individual repos from `manifest.yml` (though they'd only be set for configure and build, not at runtime).

robertmaynard:
> @robertmaynard Ah, I missed that. No, devcontainers does not change the cache location.
> 
> It seems risky for the `clean-<repo>-cpp` command to delete things outside the build dir.
> 
> Can we change the cache location to under the build dir (it's controllable via envvar, right?), so then it gets deleted along with the builddir contents. #221 adds the ability to define envvars for individual repos from `manifest.yml` (though they'd only be set for configure and build, not at runtime).

@trxcllnt Changing the cache path at build time via the env var won't be sufficient, since the tests will run without that env var and fallback to the cache under the users home directory.

We also can't change the hardcoded value at compile time, since at execution time of the binary ( say after install ) the build directory might not exist.

trxcllnt:
Ok, so can't set it at build time. Can we set it as a global envvar for the devcontainer? Something like this:
```json
// devcontainer.json
"containerEnv": {
  "CUDF_JITIFY_CACHE": "/home/coder/cudf/cpp/build/latest/jitify_cache"
}
```

robertmaynard:
> Ok, so can't set it at build time. Can we set it as a global envvar for the devcontainer? Something like this:
> 
> ```json
> // devcontainer.json
> "containerEnv": {
>   "CUDF_JITIFY_CACHE": "/home/coder/cudf/cpp/build/latest/jitify_cache"
> }
> ```

Yes that should work.

robertmaynard:
Closing this PR as the approach is going to be replaced with the env var `LIBCUDF_KERNEL_CACHE_PATH`

PR #245, opened 2024-03-14
Add some CTK12.0 dev containers for CCCL CI.

alliepiper:
We need these for new library coverage.


alliepiper:
@trxcllnt @wmaxey I don't have access to request reviews, could one of you help me get these containers built? 🙏

wmaxey:
The Windows matrix changes will definitely fail as this CTK version isn't mapped to a download link... Do we need all the different host compiler versions?

alliepiper:
> The Windows matrix changes will definitely fail as this CTK version isn't mapped to a download link.

Ah, ok. I used that version of msvc because that's what the CCCL matrix is currently using. Should we bump our 2022 version in CCCL?

> Do we need all the different host compiler versions?

For now at least. We might prune this back a bit as we decide on coverage, but the current cuda-next matrix covers 12.0 and 12.3 exhaustively. The entire test cycle is about 2 minutes so they're pretty cheap to run.

alliepiper:
Sure, I'll update this afterwards.

PR #253, opened 2024-03-19
Fix legacy-editable detection

vyasr:
Currently the devcontainers are setting all non-scikit-build-core builds to use the legacy editable setting. This is problematic for projects that are using setuptools directly (or any other backend other than scikit-build-core) without using scikit-build because legacy editable mode breaks some uses cases. In particular, we currently observe this as the breaking of dask_cudf entrypoints, which are not properly installed in legacy editable mode.

I added a new tool for the detection, but I wasn't sure how exactly we should be changing all the existing use cases so I left some notes. I will need a bit of guidance on what each of those cases should do.

### DIFF
--- a/features/src/rapids-build-utils/opt/rapids-build-utils/bin/merge-compile-commands-json.sh
+++ b/features/src/rapids-build-utils/opt/rapids-build-utils/bin/merge-compile-commands-json.sh
@@ -33,6 +33,11 @@ _list_repo_paths() {
             for ((j=0; j < ${!py_length:-0}; j+=1)); do
                 local py_sub_dir="${repo}_python_${j}_sub_dir";
                 local py_path=~/"${!repo_path:-}${!py_sub_dir:+/${!py_sub_dir}}";
+                # TODO: What are we really checking here? Do we want only
+                # packages built by scikit-build-core, or do we want all
+                # compiled Python packages (by either scikit-build or
+                # scikit-build-core) vs pure Python packages (built by
+                # setuptools)?
                 if rapids-python-uses-scikit-build-core "${py_path}"; then
                     echo "${py_path}";
                 fi
--- a/features/src/rapids-build-utils/opt/rapids-build-utils/bin/merge-compile-commands-json.sh
+++ b/features/src/rapids-build-utils/opt/rapids-build-utils/bin/merge-compile-commands-json.sh
@@ -33,12 +33,8 @@ _list_repo_paths() {
             for ((j=0; j < ${!py_length:-0}; j+=1)); do
                 local py_sub_dir="${repo}_python_${j}_sub_dir";
                 local py_path=~/"${!repo_path:-}${!py_sub_dir:+/${!py_sub_dir}}";
-                # TODO: What are we really checking here? Do we want only
-                # packages built by scikit-build-core, or do we want all
-                # compiled Python packages (by either scikit-build or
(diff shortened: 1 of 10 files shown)


vyasr:
@trxcllnt when you get a chance, I'd appreciate if you could take a look at the inline TODOs I've added. I'm not sure which of those cases should really be doing what.

features/src/rapids-build-utils/opt/rapids-build-utils/bin/tmpl/python.build.wheel.tmpl.sh
```diff
@@ -68,6 +68,7 @@ build_${PY_LIB}_python_wheel() {
         $(rapids-select-pip-wheel-args "$@")
     )";
 
+    # TODO: Should this be handling the _skbuild directory from scikit-build differently?
```
trxcllnt, line 71: No, because the path to `_skbuild` is hard-coded in scikit-build. This check here is telling `scikit-build-core` to build into `build/pip/cuda12.2/release`.

features/src/rapids-build-utils/opt/rapids-build-utils/bin/update-build-dir-links.sh
```diff
@@ -46,6 +46,7 @@ update_build_dir_links() {
             for ((j=0; j < ${!py_length:-0}; j+=1)); do
                 local py_sub_dir="${repo}_python_${j}_sub_dir";
                 local py_path=~/"${!repo_path:-}${!py_sub_dir:+/${!py_sub_dir}}";
+                # TODO: Should this also preserve the _skbuild directory for scikit-build projects?
```
trxcllnt, line 49: We don't need to do anything here, because we can't use a custom build dir unless the project uses `scikit-build-core`. If no custom build dir, no need to set up symlinks.

trxcllnt:
Was there a setuptools update that fixed things? When I was testing things, setuptools-based projects would _only_ successfully build in editable mode when `SETUPTOOLS_ENABLE_FEATURES=legacy-editable` was set.

features/src/rapids-build-utils/opt/rapids-build-utils/bin/merge-compile-commands-json.sh
```diff
@@ -33,6 +33,11 @@ _list_repo_paths() {
             for ((j=0; j < ${!py_length:-0}; j+=1)); do
                 local py_sub_dir="${repo}_python_${j}_sub_dir";
                 local py_path=~/"${!repo_path:-}${!py_sub_dir:+/${!py_sub_dir}}";
+                # TODO: What are we really checking here? Do we want only
+                # packages built by scikit-build-core, or do we want all
+                # compiled Python packages (by either scikit-build or
+                # scikit-build-core) vs pure Python packages (built by
+                # setuptools)?
```
trxcllnt, line 40: This is checking if the project was built with scikit-build-core. If so, we know its build dir, and we know its build dir has a `compile_commands.json` file in it. Then we can merge all the `compile_commands.json` files into one giant one and stuff it in the containers' home dir.

This is mostly a convenience. If a user isn't using the VSCode workspace, then clangd won't work unless it finds a `compile_commands.json` file somewhere in the directory tree from the currently-opened file up to the dir VSCode has open. Since by default that dir is `$HOME`, we write this large combined file to `~/compile_commands.json`.

PR #256, opened 2024-04-02
Add FileCheck to LLVM update-alternatives list.

wmaxey:
Provides `FileCheck` instead of `FileCheck-N` to devcontainers with the LLVM feature. This allows for use of `FileCheck` when LLVM is detectable rather than needing to parse the version of LLVM to get the name of the utility.

### DIFF
--- a/features/src/llvm/install.sh
+++ b/features/src/llvm/install.sh
@@ -41,7 +41,7 @@ echo "Installing llvm-${LLVM_VERSION} compilers and tools";
 ./llvm.sh $LLVM_VERSION ${PACKAGES:-all};
 
 if [[ "${PACKAGES:-all}" == "all" ]]; then
-    PACKAGES=("clang" "clangd" "clang++" "clang-format" "clang-tidy" "lldb" "llvm-config" "llvm-cov");
+    PACKAGES=("clang" "clangd" "clang++" "clang-format" "clang-tidy" "lldb" "llvm-config" "llvm-cov" "FileCheck");
 else
     PACKAGES=(${PACKAGES});
 fi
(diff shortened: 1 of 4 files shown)


features/src/llvm/install.sh
```diff
@@ -41,7 +41,7 @@ echo "Installing llvm-${LLVM_VERSION} compilers and tools";
 ./llvm.sh $LLVM_VERSION ${PACKAGES:-all};
 
 if [[ "${PACKAGES:-all}" == "all" ]]; then
-    PACKAGES=("clang" "clangd" "clang++" "clang-format" "clang-tidy" "lldb" "llvm-config" "llvm-cov");
+    PACKAGES=("clang" "clangd" "clang++" "clang-format" "clang-tidy" "lldb" "llvm-config" "llvm-cov" "FileCheck");
```
trxcllnt, line 44: Should the logic here be something like "if (packages == all) or (llvm-tools in packages), install FileCheck alternatives"?

wmaxey: Actually, yeah I think we want this everywhere just like `clang-format`. How do we do that?

jrhemstad: Could we just add FileCheck as one the packages to install as part of this?

https://github.com/rapidsai/devcontainers/blob/838271d0a9547aea20196c569fa8f6c44ff3c0f9/matrix.yml#L43

trxcllnt: Done

trxcllnt: @jrhemstad FileCheck is a binary that's part of the `llvm-tools` package, so we just need to install an alternative that points `/usr/bin/FileCheck -> /usr/bin/FileCheck-18`

PR #263, opened 2024-04-10
Update rmm in manifest to point to subdirectory

vyasr:
This PR is necessary for https://github.com/rapidsai/rmm/pull/1526. Unfortunately there's a circular dependency since this repo also builds all RAPIDS repos. I think the best path will be that once the rmm PR is approved and all non-pip-devcontainer CI is passing we can admin-merge that PR, then we can rerun CI on this PR and then merge that one. There will be an interim period when things are broken though.


trxcllnt:
I'm also fine merging this first if it unblocks RMM

PR #266, opened 2024-04-13
Install pre-commit for cccl devcontainers

miscco:


trxcllnt (changes requested):
LGTM, but can you also update the patch version in [cccl-dev/devcontainer-feature.json](https://github.com/rapidsai/devcontainers/blob/branch-24.06/features/src/cccl-dev/devcontainer-feature.json#L4)?

trxcllnt (approved):

PR #271, opened 2024-04-18
Add libcudf manifest for wheel build in devcontainer

msarahan:
I'm not at all sure that I've done this correctly. This is intended to support local development with a devcontainer for the work in https://github.com/rapidsai/cudf/pull/15483

### DIFF
--- a/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
+++ b/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
@@ -78,6 +78,9 @@ repos:
       sub_dir: cpp/libcudf_kafka
       depends: [cudf]
   python:
+    - name: libcudf
+      sub_dir: python/libcudf
+      depends: [cudf]
     - name: pylibcudf
       sub_dir: python/pylibcudf
       depends: [cudf]


features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
```diff
@@ -61,9 +61,12 @@ repos:
       sub_dir: cpp/libcudf_kafka
       depends: [cudf]
   python:
+    - name: libcudf
+      sub_dir: python/libcudf
+      depends: [libcudf]
```
trxcllnt, line 66:
```suggestion
      depends: [cudf]
```

features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
```diff
@@ -61,9 +61,12 @@ repos:
...
     - name: cudf
       sub_dir: python/cudf
-      depends: [cudf]
+      depends: [libcudf, cudf]
```
trxcllnt, line 69:
```suggestion
      depends: [cudf]
```

trxcllnt (changes requested):
The `depends` list in Python is for defining the source-built C++ libs that the Python project needs in order to build. Yeah, it could definitely be named better :sweat_smile:.

trxcllnt (approved):

vyasr:
Before we merge this we should make sure that we properly handle this use case in all the packages. We're going to have to adapt both the CMake and the python code in the C++ wheels just a little bit (discussed with Paul earlier today).

jameslamb:
I've revived this now that https://github.com/rapidsai/cudf/pull/15483 is just about ready to be merged.

> *We're going to have to adapt both the CMake and the python code in the C++ wheels just a little bit (discussed with Paul earlier today).*

@vyasr could you share the details for this statement? What else needs to be done?

jameslamb:
> @vyasr could you share the details for this statement? What else needs to be done?

We talked about this offline. Summarizing here.

In wheel builds in `cudf` CI, compilation of `libcudf.so` happens in the `libcudf` wheel build. That wheel contains `libcudf.so`, and libraries that need it (`cudf` and `pylibcudf`) call `libcudf.load_library()` to `dlopen` it.

Here in devcontainers, `libcudf.so` is compiled outside of any wheel-building by just directly invoking `cmake`. The `libcudf` wheel won't include it, and so `libcudf.load_library()` will not load it. That's fine... as long as `libcudf.so` is installed in a location that the linker can find when something in `cudf` / `pylibcudf` require it.

I tested this and it looks like it's working as expected. Rebuilt the CUDA 12.0 pip devcontainer from this branch, with only `cudf` (from https://github.com/rapidsai/cudf/pull/15483) mounted in.

Ran the following:

```shell
# build libcudf.so + all the Python wheels
build-cudf -j

# test that 'libcudf' is importable
python -c "import libcudf; l = libcudf.load_library(); print(l)"
# None

# test that 'pylibcudf' and 'cudf' are importable
python -c "import pylibcudf;import cudf"

# test that cudf actually works
./cudf/ci/run_cudf_pytests.sh
```

Saw all the expected things:

* `libcudf.so` was only built once, and all the wheel builds found it without rebuilding it
* `libcudf`, `cudf`, and `pylibcudf` wheels were all built successfully
* `libcudf`, `cudf`, and `pylibcudf` were all importable
* the `cudf` tests all run successfully

So I think this can be merged, and should be prior to https://github.com/rapidsai/cudf/pull/15483 being merged.

vyasr:
Thanks for documenting James.

PR #272, opened 2024-04-19
Use nproc --all to determine number of processors

wence-:
For some reason, nproc looks at the value (if any) of OMP_NUM_THREADS and OMP_THREAD_LIMIT in the environment to constrain the result. If we don't pass --all and we have one of those environment variables set, then build parallelism is incorrectly limited to nproc=1.


trxcllnt (approved):
Looks good to me!

PR #275, opened 2024-04-23
Update rust feature

wence-:
Also adds completions for bash

### DIFF
--- a/features/src/rust/install.sh
+++ b/features/src/rust/install.sh
@@ -129,7 +129,7 @@ fi
 
 echo "Installing common Rust dependencies...";
 
-rustup component add rls rust-analysis rust-src rustfmt clippy 2>&1;
+rustup component add rust-analyzer rust-src rustfmt clippy 2>&1;
 
 # Add CARGO_HOME, RUSTUP_HOME and bin directory into bashrc/zshrc files (unless disabled)
 if [ "${UPDATE_RC}" = "true" ]; then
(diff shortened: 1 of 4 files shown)


trxcllnt (approved):

trxcllnt (changes requested):
Need to bump the rust feature version ([here](https://github.com/rapidsai/devcontainers/blob/branch-24.06/features/src/rust/devcontainer-feature.json#L3)).

It also looks like the tests are failing because they can't install the `rust-analyzer` package. Do you know what's going on?

features/src/rust/install.sh
```diff
@@ -145,6 +145,10 @@ if [ "${UPDATE_RC}" = "true" ]; then
     add_etc_profile_d_script rust "$(cat .bashrc | envsubst "${vars_%,}")";
 fi
 
+# Completion
+rustup completions bash | sudo tee /usr/share/bash-completion/completions/rustup > /dev/null
+rustup completions bash cargo | sudo tee /usr/share/bash-completion/completions/cargo > /dev/null
```
trxcllnt, line 150: Don't need to use `sudo` here since we're already root when building the container.
```suggestion
rustup completions bash | tee /usr/share/bash-completion/completions/rustup > /dev/null
rustup completions bash cargo | tee /usr/share/bash-completion/completions/cargo > /dev/null
```

trxcllnt (approved):

wence-:
Thanks @trxcllnt, I see the build repos check failed due to errors in cuspatial, but in the cuspatial repo everything looks fine.

trxcllnt:
@wence- yeah, looks like a change to the `cudftestutils` that wasn't published yet. The devcontainers job in this repo builds everything from source, so we see things earlier than the devcontainer jobs in the individual repos.

PR #276, opened 2024-04-23
prepend devcontainer name with the username

trxcllnt:
This is to support the users-on-a-shared-DGX workflow.


jameslamb (approved):
Thank you for this! Based on my experience so far using devcontainers out on a shared machine, I agree it'd help to remove this source of conflicts.

PR #281, opened 2024-04-29
Add support for librmm Python build in devcontainers

vyasr:
This adds the librmm Python package to devcontainers. In devcontainer builds, this package build will be a no-op because the C++ library should be built first. Also, this package is not included in the dependency list for any other package, so it shouldn't ever really be needed. It is primarily being added to keep the manifest complete and to ensure that dependency lists will properly filter out the new dependency.


vyasr:
I expect that CI will fail here and we'll just have to merge it and then merge https://github.com/rapidsai/rmm/pull/1529/

PR #283, opened 2024-04-30
add libkvikio wheel build to manifest

msarahan:
https://github.com/rapidsai/kvikio/pull/369 moves the python wheel code one level deeper so that we add support for a dedicated libkvikio wheel. This adjustment needs to be accounted for here. Confused PR run without this change: https://github.com/rapidsai/kvikio/actions/runs/8900188332/job/24441209009#step:7:1446


trxcllnt (approved):

vyasr (changes requested):
We also need to bump the feature version here. While we're at it, I would also recommend adding the libkvikio directory since that's happening in the same kvikio PR.

features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
```diff
@@ -43,7 +43,7 @@ repos:
       depends: [rmm]
   python:
     - name: kvikio
-      sub_dir: python
+      sub_dir: python/kvikio
       depends: [KvikIO]
       args: {cmake: -DFIND_KVIKIO_CPP=ON}
```
vyasr, line 48: We no longer need this flag.
```suggestion
```

msarahan:
I don't think this PR can pass without https://github.com/rapidsai/kvikio/pull/369. @vyasr you wanted to merge this end of the loop first, right? Is it OK to merge with it failing here?

If we want things to be passing prior to merging, then I think we have to approach it on the kvikio side with the part that we ripped out here: https://github.com/rapidsai/kvikio/pull/369/files#diff-1eb4e5fd5611777d4e597ef299a1cb5ba8050c28a2dabbd4fbc56205d69e5ddaR70-R72

then merge this devcontainers PR, then remove the stuff from kvikio.

vyasr:
Yeah, one of these has to be merged while failing. Once the kvikio PR is passing all tests other than the devcontainer jobs, and once you have all the approvals that you need there, we can merge this PR and rerun tests a final time on that repo to make sure everything looks right before merging.

msarahan:
@trxcllnt updated the kvikio path from https://github.com/rapidsai/kvikio/pull/392 in https://github.com/rapidsai/devcontainers/pull/321

This PR now only adds an entry for building the cpp-only shared library wheel, which is utilized in https://github.com/rapidsai/kvikio/pull/369

msarahan:
The kvikio PR has been merged. I'm going to close/reopen this PR to cycle CI and see where we stand.

vyasr (approved):

PR #286, opened 2024-05-06
Add TBB to the windows images.

alliepiper:
This is for CCCL/Thrust.


wmaxey:
AFAICT these changes look good to me.

wmaxey:
Yeah, setting the envvar looks good. That will export the `TBB_ROOT` into the image at runtime.

alliepiper:
Perfect, that should be all we need for CMake to pick this up.

PR #289, opened 2024-05-07
Install TBB from github release.

alliepiper:
The Ubuntu 18.04 TBB packages don't include a functional CMake package. The packages from github do.


trxcllnt (approved):

alliepiper:
We don't need this. The apt version will work if I bring back our legacy `FindTBB.cmake`.

PR #291, opened 2024-05-07
libraft also requires librmm

vyasr:
The libraft entry added in #287 needs to also account for the inter-wheel dependency between libraft and librmm.

### DIFF
--- a/features/src/rapids-build-utils/devcontainer-feature.json
+++ b/features/src/rapids-build-utils/devcontainer-feature.json
@@ -1,7 +1,7 @@
 {
   "name": "NVIDIA RAPIDS devcontainer build utilities",
   "id": "rapids-build-utils",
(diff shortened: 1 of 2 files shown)


trxcllnt:
I'm not sure this is doing what you think it is.

The `depends` key is for C++ or Python projects to depend on other C++ libs. Only other C++ libs are valid entries in this list, and `librmm` looks like a Python wheel. This is so we can define the `-D<pkg>_ROOT=` arguments used to point CMake at the build dirs for other C++ projects.

Does the libraft wheel's CMake do a `find_package(librmm)` and expect to see the `librmm` scikit-build dir? If so, we'll need to modify other logic in `rapids-build-utils` to define `-D<pkg>_ROOT=` for Python libs.

PR #278, opened 2024-04-26
Test Python limited API compilation

vyasr:
This PR is leveraging the convenience of the devcontainer framework for having a long-lived test of this functionality exist in a reproducible fashion.

PR #292, opened 2024-05-08
Handle arguments with spaces in `CMAKE_ARGS`

trxcllnt:
Escape quotes so arguments with spaces are handled appropriately in the `CMAKE_ARGS` envvar.


trxcllnt:
This PR should make it unnecessary to double-quote arguments with spaces (e.g. [here](https://github.com/rapidsai/devcontainers/pull/278/files#diff-947fbd3de9fbbca870d462ec9822e48b54f7e8908123fd6a71709155b3000599R18-R92)).

vyasr (approved):

PR #322, opened 2024-06-10
resolve file_key deprecation warning from rapids-dependency-file-generator

jameslamb:
Contributes to https://github.com/rapidsai/dependency-file-generator/issues/89

Resolves this warning coming from `rapids-dependency-file-generator`.

> /opt/conda/lib/python3.9/site-packages/rapids_dependency_file_generator/cli.py:98: UserWarning: The use of --file_key is deprecated. Use -f or --file-key instead.

## Notes for Reviewers

This warning has been in there for a little over 2 months, since https://github.com/rapidsai/dependency-file-generator/pull/71 (which was in `rapids-dependency-file-generator==1.13.0`).

I think it should be fine to make this change here, as `rapids-dependency-file-generator` is reinstalled at startup.

https://github.com/rapidsai/devcontainers/blob/bff06939d44479ae1f63c3f128881d6d533de99c/features/src/rapids-build-utils/install.sh#L42

### DIFF
--- a/features/src/rapids-build-utils/devcontainer-feature.json
+++ b/features/src/rapids-build-utils/devcontainer-feature.json
@@ -1,7 +1,7 @@
 {
   "name": "NVIDIA RAPIDS devcontainer build utilities",
   "id": "rapids-build-utils",
-  "version": "24.8.3",
+  "version": "24.8.5",
   "description": "A feature to install the RAPIDS devcontainer build utilities",
   "containerEnv": {
     "BASH_ENV": "/etc/bash.bash_env"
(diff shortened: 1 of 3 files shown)


features/src/rapids-build-utils/devcontainer-feature.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "NVIDIA RAPIDS devcontainer build utilities",
   "id": "rapids-build-utils",
-  "version": "24.8.3",
+  "version": "24.8.5",
```
trxcllnt, line 4:
```suggestion
  "version": "24.8.4",
```

trxcllnt (approved):

PR #324, opened 2024-06-11
Use ubuntu24.04 for gcc-13

trxcllnt:
cc: @wmaxey


trxcllnt:
Guess we can't do this yet.

PR #325, opened 2024-06-11
support rapids-build-backend

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/31

* adds `rapids-build-backend` flags for pip builds
* updates `rapids-python-uses-scikit-build-core` to correctly detect `scikit-build-core` even when `rapids-build-backend` is the top-level build-backend

## How I tested this

Rebuilt the container, then ran the following in the CUDA 12.2 pip container.

```shell
# clean previous build outputs
uninstall-all -j
clean-all -j

# work on latest branches of everything
rapids-checkout-same-branch --omit ucxx
git -C ./ucxx checkout branch-0.39
rapids-pull-repositories

# install all the other dependencies
rapids-make-pip-env --force

# build all the packages
build-all -j

# run the cudf unit tests
cd ./cudf
./ci/run_cudf_pytests.sh

# run the cuml unit tests
cd ./cuml
./ci/run_cuml_singlegpu_pytests.sh
```

Saw that all succeed.

### DIFF
--- a/features/src/rapids-build-utils/opt/rapids-build-utils/bin/tmpl/python.build.wheel.tmpl.sh
+++ b/features/src/rapids-build-utils/opt/rapids-build-utils/bin/tmpl/python.build.wheel.tmpl.sh
@@ -82,8 +82,11 @@ build_${PY_LIB}_python_wheel() {
 
     # devcontainers builds always prefer to build projects unsuffixed, e.g. 'cudf' instead of 'cudf-cu12'
     if rapids-python-uses-rapids-build-backend "${PY_SRC}"; then
-        pip_args+=(-C "rapidsai.disable-cuda=true")
-        # pip_args+=(-C "rapidsai.matrix_entry=cuda=${CUD_VERSION}")
+        pip_args+=(-C "rapidsai.disable-cuda=true");
+        # pip_args+=(-C "rapidsai.matrix_entry=cuda=${CUDA_VERSION}")
+        exit 111;
+    elif
+        exit 112;
     fi
 
     pip_args+=("${PY_SRC}");
--- a/features/src/rapids-build-utils/opt/rapids-build-utils/bin/tmpl/python.build.wheel.tmpl.sh
+++ b/features/src/rapids-build-utils/opt/rapids-build-utils/bin/tmpl/python.build.wheel.tmpl.sh
@@ -80,11 +80,6 @@ build_${PY_LIB}_python_wheel() {
         pip_args+=(-C "build-dir=$(rapids-maybe-clean-build-dir "${cmake_args[@]}" -- "${PY_SRC}")");
     fi
 
-    # devcontainers builds always prefer to build projects unsuffixed, e.g. 'cudf' instead of 'cudf-cu12'
-    if rapids-python-uses-rapids-build-backend "${PY_SRC}"; then
-        pip_args+=(-C "rapidsai.disable-cuda=true");
-    fi
-
     pip_args+=("${PY_SRC}");
 
     trap "rm -rf '${PY_SRC}/${py_lib//"-"/"_"}.egg-info'" EXIT;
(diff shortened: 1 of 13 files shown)


features/src/rapids-build-utils/opt/rapids-build-utils/bin/tmpl/python.build.wheel.tmpl.sh
```diff
@@ -82,6 +82,11 @@ build_${PY_LIB}_python_wheel() {
         pip_args+=(-C "build-dir=$(rapids-maybe-clean-build-dir "${cmake_args[@]}" -- "${PY_SRC}")");
     fi
 
+    # devcontainers builds always prefer to build projects unsuffixed, e.g. 'cudf' instead of 'cudf-cu12'
+    if rapids-python-uses-rapids-build-backend "${PY_SRC}"; then
+        pip_args+=(-C "rapidsai.disable-cuda=true");
+    fi
+
```
trxcllnt, line 89: Is this something we could encode in the `manifest.yaml` for the projects?

trxcllnt: Though now that I say that, I am recalling that we can encode arbitrary shell scripts from entries in `manifest.yaml` as well, since they get inlined and executed in the generated script.

This makes me want to move the `scikit-build-core` parts into `manifest.yaml` as well.

jameslamb: oooooo interesting. I like the idea of not introducing another script for this and driving it from `manifest.yaml`.

Is `python.[].args.wheel` in `manifest.yaml` here supported already for that?

Like I see over in DLFW's manifest:

```yaml
python:
      - name: wholegraph
        sub_dir: python/pylibwholegraph
        args:
          wheel: *rapids_python_wheel_args_run
```

trxcllnt: Yeah exactly. There's also `args: { install: [...] }`, so we should be able to define it there

jameslamb: oh awesome! Ok I'll try that, thank you.

trxcllnt: Done

features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
```diff
@@ -3,6 +3,11 @@ x-git-defaults: &git_defaults
     tag: branch-24.08
     upstream: rapidsai
 
+x-rapids-build-backend-args: &rapids_build_backend_args |
+  --config-settings skbuild.strict-config=false
+  --config-settings rapidsai.disable-cuda=true
+  --config-settings rapidsai.matrix_entry=cuda=$(grep -o '^[0-9]*.[0-9]*' <<< "${CUDA_VERSION}")
```
jameslamb, line 9:
```suggestion
  --config-settings rapidsai.matrix_entry=cuda=$(grep -o '^[0-9]+.[0-9]+' <<< "${CUDA_VERSION}")
```

Very very minor thing, but I think we want `+` here so this will fail if `CUDA_VERSION` accidentally contains something like `"."`.

```bash
grep -o '^[0-9]*.[0-9]*' <<< "."
# .

echo $?
# 0

grep -o '^[0-9]+.[0-9]+' <<< "."
# (empty)

echo $?
# 1
```

trxcllnt:
Looks like I messed up the history trying to rebase out the commits from main. Luckily this will go away when we squash!

trxcllnt (approved):

PR #330, opened 2024-06-24
Use `--https-url` as the hostname in gh/glab API requests

trxcllnt:


trxcllnt:
Merged https://github.com/rapidsai/devcontainers/pull/331 into this PR

PR #331, opened 2024-06-24
gitlab graphql path update

cwharris:
update graphql gitlab query to return repo and repo owner path inste of display name


trxcllnt:
Closing in favor of https://github.com/trxcllnt/devcontainers/pull/14

Issue #341, opened 2024-07-03
Move to clang-format 18


bernhardmgruber:
Our CCCL pre-commit [informed us](https://github.com/NVIDIA/cccl/pull/1928) of an available update that includes moving from clang-format 17.0.6 to 18.1.8. @jrhemstad pointed out this needs to be synced with the version installed in the devcontainers, so developers will have a good experience of getting the same formatting locally as what the CI checks for. Hence this issue.

I kindly ask to upgrade the clang-format version to the latest 18 release in devcontainers. We can then fix the pre-commit clang-format version to that version.

trxcllnt:
You can bump the CCCL clang format version by updating [this line](https://github.com/rapidsai/devcontainers/blob/2f216e1a663082248a8d0708af25b0bbbf74f5e7/matrix.yml#L44) in the image matrix.

PR #343, opened 2024-07-03
Upgrade CCCL clang-format to version 18

bernhardmgruber:
Fixes: #341


trxcllnt:
@bernhardmgruber This looks good to me. Once we merge this and publish new containers, clang-format-18 could potentially produce different formatting than 17 for users. Does merging this need to be sync'd with a PR that updates/re-formats CCCL sources?

bernhardmgruber:
@trxcllnt yes, it needs to be synced with this PR in CCCL: https://github.com/NVIDIA/cccl/pull/1928. I can take care of this once you merge.

trxcllnt:
Ok, I'll merge this now and new images should be published shortly after.

PR #332, opened 2024-06-26
Update to CUDA 12.5.

bdice:
This PR updates the CUDA default to 12.5 and also adds RAPIDS devcontainers for CUDA 12.5.

Part of https://github.com/rapidsai/build-planning/issues/73.

PR #366, opened 2024-07-22
Disable CUDA driver requirement.

bdice:
This PR disables CUDA driver version compatibility checks. It doesn't seem like the container is working as intended, because driver 535 (LTS) is failing with CUDA 12.5. This is discussed further in https://github.com/rapidsai/devcontainers/pull/332.

### DIFF
--- a/features/src/cuda/devcontainer-feature.json
+++ b/features/src/cuda/devcontainer-feature.json
@@ -133,7 +133,8 @@
   "containerEnv": {
     "BASH_ENV": "/etc/bash.bash_env",
     "NVIDIA_VISIBLE_DEVICES": "all",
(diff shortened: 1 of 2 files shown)


trxcllnt (approved):

trxcllnt:
It looks like wholegraph is missing an `nvml` entry in dependencies.yaml

bdice:
Or maybe `cuda-nvml-dev`, from the log. It seems like https://github.com/rapidsai/wholegraph/pull/191 was the cause, and not this PR, so I'm going to go ahead and merge. I've opened https://github.com/rapidsai/wholegraph/pull/197 to fix this.

PR #375, opened 2024-07-30
Generate command to build pylibcudf

lithomas1:
This makes the build-pylibcudf-python command and makes build-cudf depend on it.

Potentially needs admin-merge when changes look good?
cc @vyasr


features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
```diff
@@ -78,9 +78,13 @@ repos:
...
       depends: [cudf]
   python:
+    - name: pylibcudf
+      sub_dir: python/pylibcudf
+      depends: [cudf]
+      args: {cmake: -DFIND_CUDF_CPP=ON, install: *rapids_build_backend_args}
     - name: cudf
       sub_dir: python/cudf
-      depends: [cudf]
+      depends: [cudf, pylibcudf]
```
trxcllnt, line 87: Don't need to modify this line, as this indicates a dependency on the C++ `cudf` library:

```suggestion
      depends: [cudf]
```

lithomas1:
@vyasr @jameslamb 

Can I also get a review here since y'all are already reviewing the other PR?

Thanks.

features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
```diff
@@ -78,6 +78,10 @@ repos:
       sub_dir: cpp/libcudf_kafka
       depends: [cudf]
   python:
+    - name: pylibcudf
+      sub_dir: python/pylibcudf
+      depends: [cudf]
+      args: {cmake: -DFIND_CUDF_CPP=ON, install: *rapids_build_backend_args}
     - name: cudf
       sub_dir: python/cudf
       depends: [cudf]
```
vyasr, line 87: You'll want to change this to also depend on pylibcudf.

vyasr (changes requested):

features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
bdice: https://github.com/rapidsai/devcontainers/pull/375#discussion_r1697682716

lithomas1: Yeah, I think this doesn't require any changes.

I believe build-cudf does the blocks in order on the python side.
(i.e. cudf-kafka does not depend on cudf-python here even though it needs to compile against the cudf-python Cython)

vyasr: Oh you're totally right, I forgot how this gets used. This is just to set up the CMake arguments for build dependencies, Python builds just rely on the ordering of the manifest.

vyasr:
Once CI is passing on https://github.com/rapidsai/cudf/pull/16299 for every other job, we can admin merge this one, redeploy, and rerun the devcontainer tests on that PR.

jameslamb (approved):
agree with @vyasr 's comment about the order of merging for this.

vyasr (approved):

PR #377, opened 2024-08-06
use SKBUILD_BUILD_VERBOSE instead of SKBUILD_CMAKE_VERBOSE

jameslamb:
Proposes switching from `SKBUILD_CMAKE_VERBOSE` to `SKBUILD_BUILD_VERBOSE` in wheel-building scripts in `rapids-build-utils`.

Setting `SKBUILD_CMAKE_VERBOSE` in the environment causes an unavoidable build-time exception when using `scikit-build-core>=0.10.0` and setting `minimum-version = "0.10.0"` in `scikit-build-core`.

```text
ERROR: Cannot set cmake.verbose if minimum-version is set to 0.10 or higher
```

([build link](https://github.com/rapidsai/rmm/actions/runs/10274129734/job/28430178007?pr=1637#step:7:1878))

See https://scikit-build-core.readthedocs.io/en/latest/configuration.html#verbosity

> Changed in version 0.10: cmake.verbose was renamed to build.verbose.

## Notes for Reviewers

For more context, see the conversation on that PR and https://github.com/rapidsai/build-planning/issues/58.


jakirkham (approved):

jakirkham:
Looks like the CI failures here are the same ones already seen with `scikit-build-core`

xref: https://github.com/rapidsai/rmm/pull/1638#issuecomment-2272245102

jameslamb:
> Looks like the CI failures here are the same ones already seen with `scikit-build-core`

I just restarted CI, now that there's a v0.10.1 of `scikit-build-core` up on PyPI and conda-forge which RAPIDS 24.08 builds successfully with:

* https://anaconda.org/conda-forge/scikit-build-core
* https://pypi.org/project/scikit-build-core/0.10.1/

I'm going to move this back to draft, let's not merge it yet. If those builds end up succeeding without any changes to any RAPIDS libraries, then that means we could target this at `branch-24.10` instead.

Will check back in the morning.

trxcllnt (approved):

AyodeAwe (approved):

bdice (approved):

PR #382, opened 2024-08-21
Specify use_cuda_wheels=false in matrix entry

KyleFromNVIDIA:
With packages depending on CUDA wheels by default, we want to disable that behavior in devcontainers. Add a `use_cuda_wheels=false` matrix entry.

Contributes to https://github.com/rapidsai/build-planning/issues/35


jameslamb (approved):
Perfect, thank you! I think this is totally safe to add unconditionally here, and it'll help avoid build failures as we start merging PRs like https://github.com/rapidsai/raft/pull/2415

PR #384, opened 2024-08-27
Also filter the JNI build directory for cudf

vyasr:


vyasr:
I tested this with a devcontainer where I had the JNI built and the before/after differences are as expected. In particular, because the JNI build now builds arrow, without this change all arrow libraries are considered "rapids Python conda packages" and are excluded from environments as a result. This change fixes that.

trxcllnt (approved):

PR #386, opened 2024-08-28
Add llvm-tools to all CCCL devcontainers

wmaxey:
This allows using FileCheck, hopefully from LLVM-18, in all devcontainers.

This fixes some inconsistency with how FileCheck parses nested `[[[]]]`.


wmaxey:
> I believe @trxcllnt is a better reviewer here. The only issue I'd mention is to consider how much size this adds to the image.

Uhhhh, yeah that's a good point... How large is the tools package?

PR #388, opened 2024-09-05
Pin to Python 3.11.

bdice:
Pin devcontainers to Python 3.11.

During the migration to Python 3.12, we are getting `py=3.12` as a matrix selector for rapids-dependency-file-generator, which in turn selects `python=3.12` for repositories that have added support. However, for repositories that have not migrated yet, we get the fallback matrix `python>=3.10,<3.12` which conflicts with the other pinning. To avoid this, we can pin devcontainers to use `PYTHON_VERSION="3.11"` which will be passed through to the logic here: https://github.com/rapidsai/devcontainers/blob/f9155e8267cb531efcd041eb35217b82f4083aeb/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-conda-dependencies.sh#L62-L73

### DIFF
--- a/matrix.yml
+++ b/matrix.yml
@@ -157,7 +157,7 @@ include:
   - { features: [*python, *clang_format_rapids, *clangd_dev, *cuda_curr_max_rapids, *ucx_rapids, *openmpi], env: *gcc_env }
 
   # mambaforge
-  - features: [*conda]
+  - { features: [*conda], env: { "PYTHON_VERSION": "3.11" } }
 
   ###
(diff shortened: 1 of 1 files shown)


trxcllnt (approved):

jameslamb (approved):
Thanks for looking into this!

I hadn't considered the way that Python version detection here would have been silently hitting the fallback matrices until at least one repo had a match.

I've updated https://github.com/rapidsai/build-planning/issues/40 with checklist items so when we go to add Python 3.13 in the future, we remember to do this again.

PR #397, opened 2024-09-23
Various fixes to args

vyasr:
There have been a number of changes to RAPIDS builds over the course of this release and not all changes were fully propagated to the devcontainers repo. This repo addresses the following:
- As of https://github.com/rapidsai/kvikio/pull/369, kvikio produces wheels, and https://github.com/rapidsai/kvikio/pull/439 contains critical fixes that allow the kvikio Python wheel to use the C++ libkvikio wheel. In RAPIDS Python builds we have consistently removed support for the Python build triggering the C++ build as we have created C++ wheels since in both conda and pip environments we now expect the library to be found and we do not need to automatically support the more esoteric use case of someone turning off build isolation but not having the C++ library available (devs can handle this case themselves if they wish). As a result, once https://github.com/rapidsai/kvikio/pull/466 is merged, the `FIND_KVIKIO_CPP` variable will be completely superfluous and we can remove that here.
- As of https://github.com/rapidsai/cudf/pull/16640 libcudf no longer links to libarrow and `USE_LIBARROW_FROM_PYARROW` is no longer used.
- The libcudf and libcuspatial Python package builds in the devcontainer should (like all other Python packages) omit the CUDA version suffix. For that, they need to use the `rapids_build_backend_args`.

### DIFF
--- a/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
+++ b/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
@@ -58,11 +58,11 @@ repos:
     - name: libkvikio
       sub_dir: python/libkvikio
       depends: [KvikIO]
-      args: {cmake: -DFIND_KVIKIO_CPP=ON, install: *rapids_build_backend_args}
+      args: {install: *rapids_build_backend_args}
     - name: kvikio
       sub_dir: python/kvikio
       depends: [KvikIO]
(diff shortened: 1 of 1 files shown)


vyasr:
We can take this out of draft once https://github.com/rapidsai/kvikio/pull/466 merges.

trxcllnt (approved):

PR #409, opened 2024-10-21
Fix comment about Mambaforge install

jakirkham:


jameslamb (approved):

features/src/mambaforge/install.sh
```diff
@@ -23,7 +23,7 @@ wget --no-hsts -q -O /tmp/miniforge.sh \
 
 echo "Installing Mambaforge...";
 
-# Install miniconda
+# Install Mambaforge
```
bdice, line 26: We need to replace mambaforge with miniforge everywhere in this repo.

jakirkham: Yes, but I don't want to take that on in this PR. Please raise a new issue/PR

bdice (approved):

features/src/mambaforge/install.sh
bdice: Okay. We can merge as-is for now.

PR #418, opened 2024-11-14
Remove nx-cugraph from cugraph

eriknw:
I believe this is needed to remove `nx-cugraph` from the `cugraph` repo being done in https://github.com/rapidsai/cugraph/pull/4756

This is a subset of the changes @jameslamb is making in https://github.com/rapidsai/devcontainers/pull/417. Does it make sense to do this change for `nx-cugraph` and `cugraph-gnn` independently (how close is `cugraph-gnn` to being removed?)? Should we also add `nx-cugraph` as done in that PR? Anything else?

@nv-rliu @jameslamb @rlratzel

### DIFF
--- a/features/src/rapids-build-utils/devcontainer-feature.json
+++ b/features/src/rapids-build-utils/devcontainer-feature.json
@@ -1,7 +1,7 @@
 {
   "name": "NVIDIA RAPIDS devcontainer build utilities",
   "id": "rapids-build-utils",
-  "version": "24.12.2",
+  "version": "24.12.3",
   "description": "A feature to install the RAPIDS devcontainer build utilities",
   "containerEnv": {
     "BASH_ENV": "/etc/bash.bash_env"
--- a/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
+++ b/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
@@ -240,6 +240,14 @@ repos:
       sub_dir: python/cugraph-service/server
       args: {install: *rapids_build_backend_args}
 
+- name: nx-cugraph
+  path: nx-cugraph
+  git: {<<: *git_defaults, repo: nx-cugraph}
+  python:
+    - name: nx-cugraph
+      sub_dir: .
+      args: {install: *rapids_build_backend_args}
+
 - name: cuspatial
   path: cuspatial
   git: {<<: *git_defaults, repo: cuspatial}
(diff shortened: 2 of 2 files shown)


jameslamb (changes requested):
This is not the right fix. Don't you still want `nx-cugraph` to be buildable with devcontainers? Even if we don't add devcontainer CI / devcontainer configs in the https://github.com/rapidsai/nx-cugraph repo, I think it'd still be useful to have that support here in the unified devcontainer... so you could make a change in cugraph and then test how it affects nx-cugraph.

That is why in #417, I *added* something like this:

```yaml
- name: nx-cugraph
  path: nx-cugraph
  git: {<<: *git_defaults, repo: nx-cugraph}
  python:
    - name: nx-cugraph
      sub_dir: .
      args: {install: *rapids_build_backend_args}
```

And please see the other change there... any time you change the content of something in `features/` here, you need to bump it's version.

image

Once those are addressed, I support moving this PR forward and not trying to wait on the GNN projects. I'd hoped to be able to do this all at once, but `cugraph-gnn` needs a bit more work.

jameslamb:
Also note that the CI failures here weren't because of your changes, and should resolve on a re-run now that https://github.com/rapidsai/raft/pull/2490 was merged.

eriknw:
Thanks @jameslamb I suspected we needed to make those changes, thanks for explaining :) . Updated.

trxcllnt:
> @jameslamb Even if we don't add devcontainer CI / devcontainer configs in the [nx-cugraph](https://github.com/rapidsai/) repo

We should add devcontainers to the new nx-cugraph repo. Aside from them being a convenient way to work on a project for outside contributors, they also help ensure nx-cugraph changes don't break devcontainer and DLFW builds.

jameslamb:
> We should add devcontainers to the new nx-cugraph repo

Great, let's do that. Want to put up a PR at https://github.com/rapidsai/nx-cugraph?

jameslamb:
Believe CI errors like this:

```text
/home/coder/cuvs/cpp/src/neighbors/iface/iface.hpp(176): error: incomplete type is not allowed

1 error detected in the compilation of "/home/coder/cuvs/cpp/src/neighbors/iface/iface_cagra_float_uint32_t.cu".
```

([build link](https://github.com/rapidsai/devcontainers/actions/runs/11841308604/job/32997252347?pr=418))

Should be fixed by https://github.com/rapidsai/cuvs/pull/467

jameslamb:
I put up a PR to add devcontainers CI jobs for `nx-cugraph`: https://github.com/rapidsai/nx-cugraph/pull/25

jameslamb (approved):

jameslamb:
Even though this change is also covered in #417, CI just started over there and takes a couple hours to run.

Merging this to unblock https://github.com/rapidsai/cugraph/pull/4756 and https://github.com/rapidsai/nx-cugraph/pull/25

PR #421, opened 2024-11-15
prefer system installs of UCX libraries when using RAPIDS wheels

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/118

Should only be merged if / after we merge https://github.com/rapidsai/ucx-wheels/pull/13. That PR switches `libucx` wheels' loading behavior, such that `libucx.load_library()` (which `ucx-py` and `libucxx` call at import time) prefers the `libuc{m,p,s}.so` provided by the `libucx` wheels.

https://github.com/rapidsai/ucx-wheels/pull/13 introduces an environment variable to control that... this proposes setting that, to continue preferring system-installed UCX libraries in devcontainers.

### DIFF
--- a/features/src/rapids-build-utils/devcontainer-feature.json
+++ b/features/src/rapids-build-utils/devcontainer-feature.json
@@ -4,8 +4,7 @@
   "version": "24.12.4",
   "description": "A feature to install the RAPIDS devcontainer build utilities",
   "containerEnv": {
-    "BASH_ENV": "/etc/bash.bash_env",
-    "RAPIDS_LIBUCX_PREFER_SYSTEM_LIBRARY": "true"
+    "BASH_ENV": "/etc/bash.bash_env"
   },
   "postStartCommand": [
     "/bin/bash",
--- a/features/src/rapids-build-utils/devcontainer-feature.json
+++ b/features/src/rapids-build-utils/devcontainer-feature.json
@@ -1,10 +1,11 @@
 {
   "name": "NVIDIA RAPIDS devcontainer build utilities",
   "id": "rapids-build-utils",
-  "version": "24.12.3",
+  "version": "24.12.4",
   "description": "A feature to install the RAPIDS devcontainer build utilities",
   "containerEnv": {
-    "BASH_ENV": "/etc/bash.bash_env"
+    "BASH_ENV": "/etc/bash.bash_env",
+    "RAPIDS_LIBUCX_PREFER_SYSTEM_LIBRARY": "true"
   },
   "postStartCommand": [
     "/bin/bash",
--- a/.devcontainer/rapids.Dockerfile
+++ b/.devcontainer/rapids.Dockerfile
@@ -28,6 +28,8 @@ RUN apt update -y \
 
 ENV DEFAULT_VIRTUAL_ENV=rapids
 
+ENV RAPIDS_LIBUCX_PREFER_SYSTEM_LIBRARY=true
+
 FROM ${BASE} as conda-base
 
 ENV DEFAULT_CONDA_ENV=rapids


vyasr:
@trxcllnt does this look fine to you?

features/src/rapids-build-utils/devcontainer-feature.json
```diff
@@ -1,10 +1,11 @@
 {
   "name": "NVIDIA RAPIDS devcontainer build utilities",
   "id": "rapids-build-utils",
-  "version": "24.12.3",
+  "version": "24.12.4",
   "description": "A feature to install the RAPIDS devcontainer build utilities",
   "containerEnv": {
-    "BASH_ENV": "/etc/bash.bash_env"
+    "BASH_ENV": "/etc/bash.bash_env",
+    "RAPIDS_LIBUCX_PREFER_SYSTEM_LIBRARY": "true"
```
trxcllnt, line 8: This will apply to both conda and pip containers, but I don't think that's what we want in the conda envs? 

Normally we'd put envvars that should only apply to pip devcontainers in the pip image base, e.g. [here](https://github.com/rapidsai/devcontainers/blob/7b72a8d9513af9a994182ef36b88a46d0a7d2594/.devcontainer/rapids.Dockerfile#L29) for this repo, and [here](https://github.com/rapidsai/rmm/blob/c7fc01794a78ccb6b4dc173d4ed20db156ebb39a/.devcontainer/Dockerfile#L8) for RMM etc.

trxcllnt (changes requested):
Can we put this envvar in the pip section of the respective devcontainers' Dockerfiles?

features/src/rapids-build-utils/devcontainer-feature.json
vyasr: By that logic why would I put it in the rapids.Dockerfile that you've linked to? That is the base container for both our pip and conda devcontainers, so shouldn't we put it into the devcontainer.json files in `.devcontainer/cuda*-pip`?

vyasr: Ah got it I missed that. OK cool I'll toss it in there.

vyasr: Done.

jameslamb: My thinking with putting this in `features/src/rapids-build-utils/devcontainer-feature.json` was "this environment variable only matters if you're using the `rapids-build-utils` feature, so it should only be defined as a result of pulling in that feature".

I thought that putting it in the base image would mean it'd also be defined in devcontainers that don't use the `rapids-build-utils` feature, like those used in CCCL ([code link](https://github.com/NVIDIA/cccl/blob/main/.devcontainer/cuda12.6-llvm18/devcontainer.json)).

But I see now that I must have misunderstood something about how images are built here, like that there is more than one thing that could be called "the base image". Because images like the ones CCCL is referencing (e.g. `rapidsai/devcontainers:24.12-cpp-llvm18-cuda12.6`) look like they must not have been built from `devcontainers/.devcontainer/rapids.Dockerfile`.

For example, they don't have this `DEFAULT_VIRTUAL_ENV` environment variable set.

https://github.com/rapidsai/devcontainers/blob/7b72a8d9513af9a994182ef36b88a46d0a7d2594/.devcontainer/rapids.Dockerfile#L29

```shell
docker run \
  --rm \
  --gpus 1 \
  -it rapidsai/devcontainers:24.12-cpp-llvm18-cuda12.6 \
  /bin/bash -c 'echo ${DEFAULT_VIRTUAL_ENV}'

# (empty)
```

So just saying... totally support where this ended up, but wanted to explain my reasoning for the initial changeset you saw in this PR.

jameslamb:
> Can we put this envvar in the pip section of the respective devcontainers' Dockerfiles?

I think @vyasr 's change here meets this request, @trxcllnt could you please look again?

trxcllnt (approved):

trxcllnt:
LGTM. This covers the combined devcontainer in this repo, but we should also add this to any repos that use UCX in their pip devcontainers.

jameslamb:
Alright thanks for pointing that out! I can do that. Added a note in https://github.com/rapidsai/build-planning/issues/118

jameslamb:
> we should also add this to any repos that use UCX in their pip devcontainers.

put up PRs for all of those:

* https://github.com/rapidsai/cugraph/pull/4792
* https://github.com/rapidsai/cugraph-gnn/pull/79
* https://github.com/rapidsai/cuml/pull/6149
* https://github.com/rapidsai/cuvs/pull/501
* https://github.com/rapidsai/nx-cugraph/pull/53
* https://github.com/rapidsai/raft/pull/2514

PR #414, opened 2024-11-04
Fix kvikio casing and remove unnecessary rmm dependency

vyasr:
The case-sensitive name `KvikIO` is throwing off `find_package` searches in various places.

PR #422, opened 2024-11-18
Various fixes to the manifest

vyasr:
This PR fixes a number of issues in the manifest:
- cudf's dependency on rmm was accidentally removed in #221. Since rmm is header-only I guess nobody noticed the issue since it doesn't increase build times and only shows up if you attempt to make local changes to rmm and test them in cudf.
- UCXX_ENABLE_PYTHON was removed in https://github.com/rapidsai/ucxx/pull/257
- cuml depends on cuvs as of https://github.com/rapidsai/cuml/pull/6085
- cugraph does not depend on cugraph-ops as of https://github.com/rapidsai/cugraph/pull/4744
- Many packages were relying on transitively getting rmm or raft as dependencies, which causes problems when the intermediate dependency is removed. Extra `-D<package_name>_ROOT` variables in the CMake configure don't cause any problems so I made everything explicit.


trxcllnt:
> @vyasr cudf's dependency on rmm was accidentally removed

It wasn't removed, instead the scripts added a feature to include dependencies recursively. So if  cudf depends on kvikio, and kvikio depends on rmm, then cudf gets both `-Drmm_ROOT= -Dkvikio_ROOT=`.

vyasr:
> > @vyasr cudf's dependency on rmm was accidentally removed
> 
> It wasn't removed, instead the scripts added a feature to include dependencies recursively. So if cudf depends on kvikio, and kvikio depends on rmm, then cudf gets both `-Drmm_ROOT= -Dkvikio_ROOT=`.

That seems like a good thing to do but also error-prone to rely on exclusively. I would rather follow a similar practice to how you would handle Python dependencies, C++ includes, or Python imports: if you only use something transitively then the recursive solution is good, but if you use something directly you should still include it. Otherwise you could wind up needing to make nonlocal (e.g. if raft removes a dependency that cugraph needs you'd have to trace that) or seemingly unrelated (when we remove cugraph-ops as a dependency of cugraph here, we lose every dependency of cugraph) changes.

vyasr:
Note that in this case what happened was that https://github.com/rapidsai/devcontainers/pull/414 removed the rmm dependency from kvikio because it does not in fact depend on rmm.

vyasr:
@trxcllnt what do you want to do with this PR?

trxcllnt (approved):

bdice (approved):

bdice:
> Pretty sure this needs to be `name: pylibwholegraph`:
> 
> ```yaml
>     - name: wholegraph
>       sub_dir: python/pylibwholegraph
> ```

Agreed. I updated this in e140cff.

bdice:
> Yup that's a good change (although effectively superfluous [now that wholegraph is deprecated and the repo is archived](https://github.com/rapidsai/wholegraph/), but we can remove it in a future changeset).

It's still a library -- it just exists in the cugraph-gnn repo. This entry for `pylibwholegraph` is in the `cugraph-gnn` repo, its new home.

trxcllnt:
We'll need to backport the `wholegraph -> pylibwholegraph` fix to the 24.12 branch, because it looks like the move happened in that release.

trxcllnt (approved):

trxcllnt:
Curious that this PR seems to lead to missing symbol issues...

bdice:
@trxcllnt It’s not this PR, it’s an upstream issue caused by breaking changes in https://github.com/rapidsai/rmm/pull/1722

cc: @vyasr

vyasr:
All right, I think all of the necessary upstream changes have been merged. I'm rerunning CI, things should pass this time around.

PR #432, opened 2024-12-17
Set sccache timeout

vyasr:
For local builds we don't want sccache timing out so quickly. This number matches [what we use for other CI builds](https://github.com/rapidsai/gha-tools/blob/main/tools/rapids-configure-sccache#L9).

This PR also adds some documentation of how include/exclude files are handled by `rapids-make-conda-env`.


bdice (approved):

.devcontainer/rapids.Dockerfile
```diff
@@ -2,6 +2,7 @@
 
 ARG BASE
 ARG PYTHON_PACKAGE_MANAGER=conda
+ENV SCCACHE_IDLE_TIMEOUT=32768
```
trxcllnt, line 5: Recommend lowering this to like 10-15 minutes.

The sccache server only reads AWS credentials at startup. So if the creds expire while the sccache server is still alive, it will still use them to read/write from S3.

To mitigate this, we rotate the credentials 2hrs before they're set to expire, which is hopefully enough time to finish any current builds and the server shuts down due to the idle timeout, then will read the new creds the next time it starts up again.

```suggestion
ENV SCCACHE_IDLE_TIMEOUT=900
```

trxcllnt (dismissed):
See comment

.devcontainer/rapids.Dockerfile
vyasr: Got it, thanks for the info.

vyasr:
Builds are currently blocked on https://github.com/rapidsai/cuspatial/pull/1502.

.devcontainer/rapids.Dockerfile
```diff
@@ -31,6 +30,8 @@ ENV DEFAULT_VIRTUAL_ENV=rapids
 
 ENV RAPIDS_LIBUCX_PREFER_SYSTEM_LIBRARY=true
 
+ENV SCCACHE_IDLE_TIMEOUT=900
+
```
trxcllnt, line 34: I don't think we need this one

vyasr: Why not? Don't we need this set in both containers? Environment variables set for one won't automatically propagate to the other unless the second one uses the first one as its base via `FROM`, right?

Note that I originally tried setting this for both outputs of this Dockerfile by putting this command above the first `FROM` but [that didn't work](https://github.com/rapidsai/devcontainers/actions/runs/12378502809/job/34550538935).

trxcllnt: There's only one [final build target](https://github.com/rapidsai/devcontainers/blob/71649b53d20bdefd8caec7f431ef4732b3cb6c31/.devcontainer/rapids.Dockerfile#L37-L56), so we can set common envvars there. The build target switches which target stage it extends based on the `PYTHON_PACKAGE_MANAGER` build arg.

Your [initial commit](https://github.com/rapidsai/devcontainers/pull/432/commits/40e5a1cc8b6ce739eeb569532b7bbea88bc2acc4) only set the envvar in the conda build stage, so that's why it only showed up there. We can instead define it once at the bottom, with all the other common sccache envvars.

.devcontainer/rapids.Dockerfile
```diff
@@ -30,6 +30,8 @@ ENV DEFAULT_VIRTUAL_ENV=rapids
...
```
trxcllnt, line 34:
```suggestion
```

.devcontainer/rapids.Dockerfile
vyasr: Ohhh nice I missed that there was a final stage there dynamically based off one of the other two. OK that makes a lot of sense then.

trxcllnt:
It looks like there's a discrepancy between RAPIDS libs. Some repo wants sphinx 6 and another wants 8. Is this being tracked anywhere @vyasr?

vyasr:
> It looks like there's a discrepancy between RAPIDS libs. Some repo wants sphinx 6 and another wants 8. Is this being tracked anywhere @vyasr?

Hmm I just reran assuming that it was a transient failure and I didn't realize that this bump occurred. Let me see if I can find it.

vyasr:
I'm not seeing it in any repo's dependencies.yaml...

jameslamb:
just re-triggered CI here now that the sphinx pins were removed:

* https://github.com/rapidsai/cugraph/pull/4839
* https://github.com/rapidsai/nx-cugraph/pull/60
* https://github.com/rapidsai/cuspatial/pull/1507
* https://github.com/rapidsai/cuml/pull/6195

vyasr:
@jameslamb you'll also need to get [cugraph-ops](https://github.com/rapidsai/cugraph-ops/blob/branch-25.02/dependencies.yaml#L330) updated. I missed that one in my pass because I've already removed it from my local devcontainer setup and I forgot that we haven't quite removed it altogether yet.

jameslamb:
Ah ok, put up https://github.com/rapidsai/cugraph-ops/pull/708

jameslamb:
restarted all CI here now that the `cugraph-ops` PR is merged (thanks @bdice )

trxcllnt:
Still seems to be failing due to incompatible `breathe` version:
```
  LibMambaUnsatisfiableError: Encountered problems while solving:
    - package breathe-4.35.0-pyhd8ed1ab_0 requires sphinx >=4.0,<6.0.0a, but none of the providers can be installed
```

jameslamb:
blegh.

Looking at the logs for `conda` jobs:

```text
channels:
- rapidsai
- rapidsai-nightly
- dask/label/dev
- pytorch
- pyg
- conda-forge
- nvidia
dependencies:
...
- breathe
- breathe>=4.35.0
- sphinx
- sphinx-autobuild
- sphinx-click
- sphinx-copybutton
- sphinx-markdown-tables
- sphinx-remove-toctrees
- sphinx>=8.0.0
...
- pip:
  ...
  - breathe>=4.35.0
  - nvidia-sphinx-theme
```

([build link](https://github.com/rapidsai/devcontainers/actions/runs/12426233105/job/34812615598?pr=432))

The latest `breathe` on `conda-forge` has a ceiling of `sphinx >=4.0,<7.2`.

So we need one of these 2 things to happen:

* this to get merged: https://github.com/conda-forge/breathe-feedstock/pull/64
* every RAPIDS repo with the `breathe` conda package to switch to `pip: breathe` (and hope that no other dependencies pull it in transitively)

Or, if we could live with not supporting docs builds in devcontainers for a bit (I have no idea if anyone uses devcontainers to build docs), we could add `breathe` to the list of exclusions here, until that conda-forge PR is updated?

https://github.com/rapidsai/devcontainers/blob/71649b53d20bdefd8caec7f431ef4732b3cb6c31/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-conda-dependencies.sh#L161-L170

bdice:
> Or, if we could live with not supporting docs builds in devcontainers for a bit (I have no idea if anyone uses devcontainers to build docs), we could add breathe to the list of exclusions here, until that conda-forge PR is updated?

I am okay with this for a short-term solution as long as we add comments explaining it. I do use devcontainers to build docs, but it's rare for me. We do need to get that conda-forge PR merged to fix cuVS docs builds so they're not relying on `pip: breathe` ...

trxcllnt:
@jameslamb if you want to exclude breathe, we can do it as an argument at the callsite rather than inside this file. These exclusions are special, since they are either patterns or transform the dependency somehow.

The change I'd suggest is replacing this line with the one below:
https://github.com/rapidsai/devcontainers/blob/71649b53d20bdefd8caec7f431ef4732b3cb6c31/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-conda-env.sh#L38
```bash
    rapids-make-conda-dependencies --exclude <(echo breathe) "${OPTS[@]}" > "${new_env_path}";
```

jameslamb:
Thanks @trxcllnt .

It originally looked to me, based on these lines, that the docs for these `--include` / `--exclude` arguments were wrong, and that those really referred to *packages*, not *files*.

https://github.com/rapidsai/devcontainers/blob/71649b53d20bdefd8caec7f431ef4732b3cb6c31/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-conda-dependencies.sh#L12

https://github.com/rapidsai/devcontainers/blob/71649b53d20bdefd8caec7f431ef4732b3cb6c31/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-conda-dependencies.sh#L49-L52

https://github.com/rapidsai/devcontainers/blob/71649b53d20bdefd8caec7f431ef4732b3cb6c31/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-conda-dependencies.sh#L163-L170

That led me to https://github.com/rapidsai/devcontainers/commit/b55f3668eb10156ff5222bf57cfa66d8d67e0ed8 .

But I see now that those `-f` being appended aren't for e.g. `conda-merge`, but for like `git grep -E -f {file}`, and I guess the `<(echo breathe)` thing is to provide that single pattern as a file descriptor without touching the filesystem, like `git grep -E -f <(echo breathe)`. Pretty powerful!!!

I just pushed https://github.com/rapidsai/devcontainers/pull/432/commits/7f93231a9510258ec4b93e05e01e266d489d281e reverting my misguided changes, but including comments and expanding the docs to make it clearer that this is how that's working.

jameslamb:
It looks to me like skipping `breathe` is sufficient to get CI working here again. If we merge this as-is, I'll open an issue tracking the work to revert that `breathe` exclusion whenever https://github.com/conda-forge/breathe-feedstock/pull/64 is merged.

@trxcllnt whenever you have time, can you re-review?

bdice:
https://github.com/conda-forge/breathe-feedstock/pull/64 has been merged, so let’s try reverting the breathe-related changes.

vyasr:
lol I just came here to say the same. I'll update.

features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-conda-env.sh
```diff
@@ -35,7 +35,8 @@ make_conda_env() {
     local -r new_env_path="$(realpath -m "/tmp/${env_file_name}")";
     local -r old_env_path="$(realpath -m "${HOME}/.conda/envs/${env_file_name}")";
 
-    rapids-make-conda-dependencies "${OPTS[@]}" > "${new_env_path}";
+    # the '--exclude <(echo breathe)' can be removed when https://github.com/conda-forge/breathe-feedstock/pull/64 is merged
```
bdice, line 38:
```suggestion
```

features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-conda-env.sh
```diff
@@ -35,7 +35,8 @@ make_conda_env() {
...
+    rapids-make-conda-dependencies --exclude <(echo breathe) "${OPTS[@]}" > "${new_env_path}";
```
bdice, line 39:
```suggestion
    rapids-make-conda-dependencies "${OPTS[@]}" > "${new_env_path}";
```

jameslamb:
Can we please keep the docs and comments I'd added? They wouldn't saved me some time investigating this and I think they'll help the next person looking at something like this.

I really really don't think it's obvious that `--exclude` / `--include` can also take file descriptors, or that the those files are being treated as files of *patterns* for `grep`.

vyasr:
Sure I'll just remove my commits and make the change Bradley suggested instead then.

We need https://github.com/conda-forge/breathe-feedstock/pull/66 merged to get new Breathe packages up before CI will pass here.

vyasr:
Great we're finally in the clear on the Breathe front here. The conda environments solved on the latest runs.

vyasr:
I expect that the builds here will still fail though because I started merging rapids-logger changes :face_exhaling:

PR #436, opened 2025-01-13
Update Windows containers

wmaxey:
New packages, new compilers.

Changes how environment is prepared on the container. Skips environment initialization by embedding the msvc vars into the global env.


wmaxey:
These failures look unrelated maybe?

trxcllnt:
Yep, failures are unrelated. Windows images look like they built.

PR #439, opened 2025-01-16
Disable CUDA 12.0 RAPIDS builds.

bdice:
CUDA 12.0 RAPIDS builds appear to be blocked due to a failure caused by https://github.com/rapidsai/cudf/pull/17289.

CUDA 12.5 builds are seemingly unaffected. For now, I am opening this PR to unblock devcontainers CI.

PR #447, opened 2025-01-24
Revert "Disable CUDA 12.0 RAPIDS builds."

vyasr:
Reverts rapidsai/devcontainers#439

We had disabled these builds due to seg faults that were observed after https://github.com/rapidsai/cudf/pull/17289 was merged. Those should have been fixed after https://github.com/NVIDIA/cuCollections/pull/660 and https://github.com/rapidsai/rapids-cmake/pull/744. https://github.com/rapidsai/cudf/pull/17758 is also relevant but not a direct source for the fix.

### DIFF
--- a/.github/workflows/build-all-rapids-repos.yml
+++ b/.github/workflows/build-all-rapids-repos.yml
@@ -36,10 +36,7 @@ jobs:
       pull-requests: read
     with:
       arch: '["amd64"]'
-      # Disabling CUDA 12.0 due to failures in cudf builds. See:
-      # https://github.com/rapidsai/devcontainers/pull/438#issuecomment-2596318879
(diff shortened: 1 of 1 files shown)


vyasr:
In the interim between #439 and this PR, we merged updates across RAPIDS for https://github.com/rapidsai/build-planning/issues/129 updating to require gcc 13 for builds. CUDA 12.0 requires gcc 12 (gcc 13 support was added in CUDA 12.4). After some offline discussion, we decided that supporting CUDA 12.0 conda builds is not worth the effort of bifurcating our build environments here, so I'm going to close this PR and broadcast more publicly that we are no longer supporting building RAPIDS with CUDA 12.0 (builds with later CUDA versions will still run just fine in CUDA 12 environments, which is more important).

PR #448, opened 2025-01-26
Switch to miniforge

trxcllnt:
[Mambaforge is dead](https://conda-forge.org/news/2024/07/29/sunsetting-mambaforge/), long live Miniforge!


bdice (approved):
We should be able to use miniforge always. That is the recommended migration path away from mambaforge. Miniforge ships with mamba and has for a while, afaik.

PR #453, opened 2025-01-27
Add CUDA 12.8 to CUDA feature

bdice:
Adds 12.8 and 12.6 to the CUDA feature.

xref: https://github.com/rapidsai/build-planning/issues/139


trxcllnt (changes requested):
Should we update the default from 12.5 to 12.8? If so, will probably need to update tests.

bdice:
I was going to update the default to 12.8 later -- we want to roll out 12.8 support across RAPIDS first, I think.

jakirkham:
Agreed. To help us remember, filed a tracking issue: https://github.com/rapidsai/devcontainers/issues/455

trxcllnt (approved):

PR #463, opened 2025-01-30
Ensure that the TBB path is set correctly on Windows

alliepiper:
TBB_ROOT was not defined when configuring the PATH.


alliepiper:
@trxcllnt this should fix the thrust windows tests

PR #465, opened 2025-02-04
Use CUDA 12.8 by default in RAPIDS and CUDA features.

bdice:
Now that RAPIDS has fully migrated to CUDA 12.8, we can change the defaults in our devcontainers.

### DIFF
--- a/.github/workflows/build-all-rapids-repos.yml
+++ b/.github/workflows/build-all-rapids-repos.yml
@@ -36,7 +36,7 @@ jobs:
       pull-requests: read
     with:
       arch: '["amd64"]'
-      cuda: '["12.5"]'
+      cuda: '["12.8"]'
       node_type: cpu32
       extra-repo-deploy-key: CUMLPRIMS_SSH_PRIVATE_DEPLOY_KEY
       build_command: |
--- a/features/src/cuda/README.md
+++ b/features/src/cuda/README.md
@@ -15,7 +15,7 @@ A feature to install the NVIDIA CUDA Toolkit
 
 | Options Id | Description | Type | Default Value |
 |-----|-----|-----|-----|
-| version | Version of the CUDA Toolkit to install. | string | 12.5 |
+| version | Version of the CUDA Toolkit to install. | string | 12.8 |
 | cuDNNVersion | Version of cuDNN to install. | string | 8 |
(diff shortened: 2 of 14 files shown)


trxcllnt:
Also need to update the versions in the  cuda feature tests [here](https://github.com/rapidsai/devcontainers/blob/branch-25.02/features/test/cuda/test.sh#L20-L24)

jakirkham:
It seems the build just stops at some point. Do we have enough resources for this CI job?

trxcllnt:
@bdice I'm not sure either, but seems like a change in mamba. We symlink `ln -s /opt/conda/etc/profile.d/mamba.sh /etc/profile.d/mamba.sh` in the feature, but from this message it sounds like that's going to be removed?

```
WARNING: This message originates from /opt/conda/etc/profile.d/mamba.sh
WARNING: /opt/conda/etc/profile.d/mamba.sh (the file emitting this warning) is deprecated.
WARNING: This file will be removed in mamba 2.1.0.
WARNING: Please use 'mamba shell init' to get the correct shell scripts for your shell.
WARNING: The MAMBA_ROOT_PREFIX environment variable is not set.
WARNING: This is required for mamba to work correctly as of 2.0.
WARNING: 
WARNING: For now, we are setting 'MAMBA_ROOT_PREFIX' to '/opt/conda'.
WARNING: 
WARNING: Please make sure this is consistent with your installation or alternatively (by order of preference):
WARNING:   - rerun 'mamba shell init -s posix' to get the correct value
WARNING:   - manually set 'MAMBA_ROOT_PREFIX' to the root of your installation in your shell profile script.
WARNING:   - use the '-r,--root-prefix' CLI option when calling mamba.
WARNING: 
WARNING: This message originates from /opt/conda/etc/profile.d/mamba.sh
mkdir: cannot create directory '/opt/conda/etc/profile.d/mamba.sh': Not a directory
mkdir: cannot create directory '/opt/conda/etc/profile.d/mamba.sh': Not a directory
```

I'm building locally to see what's going on.

trxcllnt:
Oh I see the issue. The contents of `mamba.sh` have changed and it's breaking our conda shell init workarounds.

trxcllnt:
Should be fixed now

PR #470, opened 2025-02-22
Remove CUDA 12.5 devcontainers.

bdice:
Now that RAPIDS has fully migrated to CUDA 12.8, we can remove the CUDA 12.5 devcontainers.


trxcllnt:
@bdice are we tracking these linker errors anywhere? I am also seeing them in #472

PR #487, opened 2025-04-08
Misc shell script clean up

trxcllnt:
PR with miscellaneous shell script cleanup that doesn't feel right including in other PRs. The commit history is fairly self-explanatory.

The most intrusive one is around hiding variable values from xtrace. Doing `! test -n "${foo:+x}"` instead of `test -z "${foo:-}"` matters if `$foo` is sensitive and xtrace is enabled in an environment where the sensitive value isn't redacted, because the former prints `test -n x` and the later prints `test -z fooVal`.

### DIFF
--- a/features/src/cccl-dev/devcontainer-feature.json
+++ b/features/src/cccl-dev/devcontainer-feature.json
@@ -1,7 +1,7 @@
 {
   "name": "NVIDIA CCCL development utilities",
   "id": "cccl-dev",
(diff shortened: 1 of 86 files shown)


bdice (approved):
Curious — did you automate this? Did you use shellcheck or some other tool to find the things that needed to be changed?

trxcllnt:
@bdice no, this is all elbow grease

PR #488, opened 2025-04-10
Add CTK 12.0 + MSVC 14.42 container for CCCL/cudax.

alliepiper:


alliepiper:
Closing. CTK 12.0 doesn't recognize 14.43 as MSVC 2022:

```
C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.0\include\crt/host_config.h(153): fatal error C1189: #error:  -- unsupported Microsoft Visual Studio version! Only the versions between 2017 and 2022 (inclusive) are supported! The nvcc flag '-allow-unsupported-compiler' can be used to override this version check; however, using an unsupported host compiler may cause compilation failure or incorrect run time execution. Use at your own risk.
```

PR #493, opened 2025-04-18
Normalize ucxx branch name

bdice:
This PR normalizes `ucxx` branch names. When running `rapids-checkout-same-branch`, `ucxx` is always ignored. Then, if a branch name matching `branch-YY.MM` is selected, `ucxx` will use the corresponding branch name for its matching version.

### DIFF
--- a/features/src/rapids-build-utils/devcontainer-feature.json
+++ b/features/src/rapids-build-utils/devcontainer-feature.json
@@ -1,7 +1,7 @@
 {
   "name": "NVIDIA RAPIDS devcontainer build utilities",
   "id": "rapids-build-utils",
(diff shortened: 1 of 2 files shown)


bdice:
I verified locally that this works!

vyasr (approved):
Thanks Bradley! This solution looks good to me. The one edge case is if someone explicitly requests a ucx-style branch version, but I think it's fine for that to be off since you can always explicitly check out those repos in that case.

bdice:
Thanks. I think requesting ucx-version branches is out of scope for this utility, because it'd have to match the rest of RAPIDS.

PR #495, opened 2025-04-22
Build shared `libpython` in pip base images

trxcllnt:
This should fix the libucxx [link error](https://github.com/rapidsai/ucxx/actions/runs/14598480223/job/40961661967#step:9:2494) that cropped up:
```
  /usr/bin/ld: /usr/local/python/3.12.10/lib/libpython3.12.a(import.o): relocation R_X86_64_TPOFF32 against hidden symbol `pkgcontext' can not be used when making a shared object
```

### DIFF
--- a/features/src/cccl-dev/devcontainer-feature.json
+++ b/features/src/cccl-dev/devcontainer-feature.json
@@ -1,7 +1,7 @@
 {
   "name": "NVIDIA CCCL development utilities",
   "id": "cccl-dev",
-  "version": "25.6.1",
+  "version": "25.6.2",
   "description": "A feature to install NVIDIA CCCL development utilities",
   "options": {
     "litVersion": {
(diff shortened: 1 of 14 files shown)


AyodeAwe (approved):

trxcllnt:
Alternatively, maybe we need to set `"enableShared": true` for the Python feature.

bdice:
> Alternatively, maybe we need to set `"enableShared": true` for the Python feature.

That might be helpful. Can we try that?

trxcllnt:
@bdice yep, trying in https://github.com/rapidsai/devcontainers/pull/495/commits/e6356abe3114d5dccde63276924441358efd73d5

trxcllnt:
@bdice can confirm the `enableShared: true` option fixes the ucxx wheel link error :tada:

bdice (approved):
Re-approving the new changes. Let's see if CI passes before merging? Or if you feel comfortable with your local testing, feel free to move it forward.

PR #496, opened 2025-04-23
fix `rapids-python-conda-pkg-names` script for rattler recipes

trxcllnt:


bdice (approved):

wence-:
This looks correct to me. the nvtx package is now being pulled in by RMM using version 3.1.0 not being found in the conda env: https://github.com/rapidsai/devcontainers/actions/runs/14621389663/job/41022109436?pr=496#step:9:3121

PR #515, opened 2025-05-19
Drop cuGraph-DGL from Manifest

alexbarghi-nv:
Drops cuGraph-DGL from the manifest since we are no longer publishing it.


vyasr:
Holding off on approving this until the cugraph-gnn PR is merged.

trxcllnt (changes requested):
need to bump the version in `features/src/rapids-build-utils/devcontainer-feature.json`

vyasr:
> need to bump the version in `features/src/rapids-build-utils/devcontainer-feature.json`

Done.

PR #516, opened 2025-05-19
Add gcc-15 without using ubuntu25.

wmaxey:
GCC-15 should just work out of the box in Ubuntu-24 with the test PPA enabled.

### DIFF
--- a/matrix.yml
+++ b/matrix.yml
@@ -14,7 +14,7 @@ x-gcc-11: &gcc_11 { name: "gcc", version: "11" }
 x-gcc-12: &gcc_12 { name: "gcc", version: "12" }
 x-gcc-13: &gcc_13 { name: "gcc", version: "13" }
 x-gcc-14: &gcc_14 { name: "gcc", version: "14" }
-x-gcc-14: &gcc_15 { name: "gcc", version: "15" }
+x-gcc-15: &gcc_15 { name: "gcc", version: "15" }
 x-gcc-env: &gcc_env { CC: "gcc", CXX: "g++", CUDAHOSTCXX: "g++" }
 x-gcc-rapids: &gcc_rapids { name: "gcc", version: "14", hide: true }
 x-gcc-env-rapids: &gcc_env_rapids { CC: "gcc", CXX: "g++", CUDAHOSTCXX: "g++", PYTHON_VERSION: "3.14" }
--- a/matrix.yml
+++ b/matrix.yml
@@ -14,6 +14,7 @@ x-gcc-11: &gcc_11 { name: "gcc", version: "11" }
 x-gcc-12: &gcc_12 { name: "gcc", version: "12" }
 x-gcc-13: &gcc_13 { name: "gcc", version: "13" }
 x-gcc-14: &gcc_14 { name: "gcc", version: "14" }
+x-gcc-14: &gcc_15 { name: "gcc", version: "15" }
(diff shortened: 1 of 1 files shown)


trxcllnt (changes requested):
need to bump the version in `features/src/gcc/devcontainer-feature.json`

wmaxey:
> need to bump the version in `features/src/gcc/devcontainer-feature.json`

To what? `25.8.1`? Sorry I'm unfamiliar with the process on these files and when looking at the blame any updates seem to match the branch version.

trxcllnt (approved):

features/src/gcc/install.sh
```diff
@@ -23,6 +23,11 @@ check_packages                  \
 GCC_VERSION_DEFAULT="$(gcc -dumpversion)";
 GCC_VERSION="${VERSION:-${GCC_VERSION_DEFAULT}}";
 
+if [[ "${GCC_VERSION}" == "15" ]]; then
+    apt-add-repository -y "deb https://ppa.launchpadcontent.net/ubuntu-toolchain-r/ppa/ubuntu plucky main";
+    apt-add-repository -y "deb https://ppa.launchpadcontent.net/ubuntu-toolchain-r/test/ubuntu plucky main";
+fi
```
trxcllnt, line 29: Is this worth future-proofing?

```suggestion
if [[ "${GCC_VERSION}" == "15" ]]; then
    VERSION_CODENAME="$(. /etc/os-release; echo ${VERSION_CODENAME})"
    apt-add-repository -y "deb https://ppa.launchpadcontent.net/ubuntu-toolchain-r/ppa/ubuntu $VERSION_CODENAME main";
    apt-add-repository -y "deb https://ppa.launchpadcontent.net/ubuntu-toolchain-r/test/ubuntu $VERSION_CODENAME main";
fi
```

Or maybe doing something more advanced, like checking if the desired `$GCC_VERSION` is installable via `apt-cache policy` and conditionally adding the repo?

features/src/gcc/install.sh
```diff
@@ -23,6 +23,11 @@ check_packages                  \
...
+
 if [[ "${GCC_VERSION}" != "${GCC_VERSION_DEFAULT}" ]]; then
     apt-add-repository -y ppa:ubuntu-toolchain-r/test;
     apt-add-repository -y ppa:ubuntu-toolchain-r/ppa;
```
trxcllnt, line 33: @wmaxey I'm curious why these aren't sufficient? Shouldn't these lines add the PPAs for the current Ubuntu release?

trxcllnt (changes requested):
Changing back to request changes due to more questions.

features/src/gcc/install.sh
wmaxey: So, last time we just built using `23.04` and this time I was imagining it would be simpler to build by fetching from `plucky` the codename of the next release. When 26.04 comes around this would need to be removed. Either way it's kind of a hack.

Honestly I don't have good judgement for which way is better. Build with or pull from 25.04 to get gcc-15.

features/src/gcc/install.sh
wmaxey: Comment above is also in response to this.

features/src/gcc/install.sh
trxcllnt: What if we build the `gcc-15` images with ubuntu24.04? Would we need the changes here? Is there any reason we shouldn't?

trxcllnt (approved):

PR #519, opened 2025-05-23
Update rapids-dependency-file-generator

bdice:
I am running into problems solving the conda environment in https://github.com/rapidsai/rapidsmpf/pull/301 because the `rapids-dependency-file-generator` is too old. This updates it.

### DIFF
--- a/features/src/rapids-build-utils/devcontainer-feature.json
+++ b/features/src/rapids-build-utils/devcontainer-feature.json
@@ -1,7 +1,7 @@
 {
   "name": "NVIDIA RAPIDS devcontainer build utilities",
   "id": "rapids-build-utils",
-  "version": "25.8.2",
+  "version": "25.8.3",
   "description": "A feature to install the RAPIDS devcontainer build utilities",
   "containerEnv": {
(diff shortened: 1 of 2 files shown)


bdice:
This was pinned in https://github.com/rapidsai/devcontainers/pull/344 / https://github.com/rapidsai/devcontainers/pull/361 but no reason for the pinning was given.

jameslamb (approved):
> _no reason for the pinning was given_

Based on the timing and the specific version in the ceiling, I *suspect* it was a result of https://github.com/rapidsai/dependency-file-generator/pull/100

All RAPIDS repos were updated to account for that in 24.08 (including this repo, in #322, as part of https://github.com/rapidsai/dependency-file-generator/issues/89), but maybe the pin was added because there were people using newer devcontainers with older RAPIDS versions.

If I'm right about why this pin was here, it should be fine to remove it now... I'd be surprised to learn anyone would need to use the 25.08 devcontainers to build RAPIDS 24.06 or older.

trxcllnt (approved):

PR #522, opened 2025-05-29
Drop CUDA 11 devcontainers

bdice:
xref: https://github.com/rapidsai/build-planning/issues/184

### DIFF
--- a/matrix.yml
+++ b/matrix.yml
@@ -142,7 +142,8 @@ include:
   - { features: [*python_rapids, *clang_format_rapids, *clangd_dev, *cuda_curr_min], env: *gcc_env_rapids }
   - { features: [*python_rapids, *clang_format_rapids, *clangd_dev, *cuda_curr_max], env: *gcc_env_rapids }
   - { features: [*python_rapids, *clang_format_rapids, *clangd_dev, *cuda_curr_max_rapids], env: *gcc_env_rapids }
-  - { features: [*python_rapids, *clang_format_rapids, *clangd_dev, *cuda_prev_max_rapids, *ucx_rapids, *openmpi], env: *gcc_env_rapids }
+  # Commented as RAPIDS only supports one major version (CUDA 12) right now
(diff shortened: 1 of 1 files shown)


trxcllnt (approved):

matrix.yml
```diff
@@ -2,8 +2,9 @@ x-cuda-prev-min: &cuda_prev_min { name: "cuda", version: "12.0" }
 x-cuda-prev-max: &cuda_prev_max { name: "cuda", version: "12.9" }
 x-cuda-curr-min: &cuda_curr_min { name: "cuda", version: "12.0" }
 x-cuda-curr-max: &cuda_curr_max { name: "cuda", version: "12.9" }
-x-cuda-curr-min: &cuda_cccl_python { name: "cuda", version: "12.8" }
-x-cuda-prev-max-rapids: &cuda_prev_max_rapids { name: "cuda", version: "11.8" }
+x-cuda-cccl-python: &cuda_cccl_python { name: "cuda", version: "12.8" }
```
bdice, line 5: cc: @alliepiper for this change. The `x-key` was not unique. I don't think it was harming anything but it's not good practice to have duplicate keys even if the key itself is meaningless.

matrix.yml
```diff
@@ -2,6 +2,7 @@ x-cuda-prev-min: &cuda_prev_min { name: "cuda", version: "12.0" }
 x-cuda-prev-max: &cuda_prev_max { name: "cuda", version: "12.9" }
 x-cuda-curr-min: &cuda_curr_min { name: "cuda", version: "12.0" }
 x-cuda-curr-max: &cuda_curr_max { name: "cuda", version: "12.9" }
+x-cuda-curr-min: &cuda_cccl_python { name: "cuda", version: "12.8" }
```
trxcllnt:
```suggestion
x-cuda-cccl-python: &cuda_cccl_python { name: "cuda", version: "12.8" }
```

PR #527, opened 2025-06-04
Use compression=zstd for faster image builds and loads (round two)

trxcllnt:
Reopening this PR now that auth issues are worked out. Release job in this repo succeeds: https://github.com/rapidsai/devcontainers/actions/runs/15448178834


ajschmidt8:
@trxcllnt, are there any GHAs jobs using these containers that we can check to see if zstd compression resulted in notably faster decompression times?

trxcllnt:
@ajschmidt8 All the devcontainer jobs (e.g. [here](https://github.com/rapidsai/cudf/actions/runs/15520352327/job/43698798493)) in the RAPIDS repos use these images. All the CCCL PR jobs use these images as well.

bdice:
@trxcllnt Do we need to enable zstd compression in our `rapidsai/ci-imgs` and `rapidsai/docker` repos too? Is this just a devcontainers thing or a generally good practice for Docker images?

PR #528, opened 2025-06-05
Add pbzip2, more CTK coverage.

alliepiper:
pbzip2 significantly speeds up comp/decomp for large archives.

Python needs more minor versions of CTK to test.


alliepiper:
RAPIDS failures are unrelated, this only affects cccl containers.

PR #529, opened 2025-06-16
Reintroduce verbose flag to build-all in rapids repo build

wence-:
Without this, we can't see the compile commands being executed in the nightly build jobs (and hence can't debug why a local devcontainer build might be missing the sccache).


trxcllnt:
Yeah, I started a PR here: https://github.com/rapidsai/shared-workflows/pull/389

PR #530, opened 2025-06-16
Add `cu++filt` package to cccl devcontainers

davebayer:
We would like to use `libcufilt.a` library from CTK in cccl. This PR adds this library to the devcontainers

### DIFF
--- a/features/src/cuda/devcontainer-feature.json
+++ b/features/src/cuda/devcontainer-feature.json
@@ -117,11 +117,6 @@
       "default": false,
       "description": "Install CUDA Direct Linear Solvers Library (cuSOLVER)"
     },
-    "installCUXXFILT": {
-      "type": "boolean",
-      "default": false,
-      "description": "Install CUDA C++ demangler (cu++filt)"
-    },
     "installNPP": {
       "type": "boolean",
       "default": false,
--- a/features/src/cuda/devcontainer-feature.json
+++ b/features/src/cuda/devcontainer-feature.json
@@ -1,7 +1,7 @@
 {
   "name": "CUDA Toolkit",
   "id": "cuda",
-  "version": "25.8.0",
+  "version": "25.8.1",
   "description": "A feature to install the NVIDIA CUDA Toolkit",
   "options": {
     "version": {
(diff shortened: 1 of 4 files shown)


trxcllnt:
The `cuda-compiler-12-9` package should pull in `cu++filt` if it's available. Since `installCompilers: true` is the default and I don't see it set to false in `matrix.yml`, I'd expect `libcufilt.a` to already be in the CCCL images.

trxcllnt:
Oh I see. The issue is that `pruneStaticLibs` defaults to true, so `libcufilt.a` is being pruned from the image. So the logic should really be to either set `pruneStaticLibs: false` in matrix.yml, or to add an option to not delete `libcufilt.a` from the image even if `pruneStaticLibs` is true.

davebayer:
> Oh I see. The issue is that `pruneStaticLibs` defaults to true, so `libcufilt.a` is being pruned from the image. So the logic should really be to either set `pruneStaticLibs: false` in matrix.yml, or to add an option to not delete `libcufilt.a` from the image even if `pruneStaticLibs` is true.

Can I just add `libcufilt.a` among libraries that are not removed by the `prune-static-libs.sh` script, or do you really want to avoid keeping the library unless necessary?

Before:
```bash
for dir in "lib" "lib64"; do
    find "$(realpath -m "${CUDA_HOME}/${dir}")/" -type f \
        \( -name '*.a' ! -name 'libnvptxcompiler_static.a' ! -name 'libcudart_static.a' ! -name 'libcudadevrt.a' ! -name 'libculibos.a' \) \
        -delete || true;
done
```

After:
```bash
for dir in "lib" "lib64"; do
    find "$(realpath -m "${CUDA_HOME}/${dir}")/" -type f \
        \( -name '*.a' ! -name 'libnvptxcompiler_static.a' ! -name 'libcudart_static.a' ! -name 'libcudadevrt.a' ! -name 'libculibos.a' ! -name 'libcufilt.a' \) \
        -delete || true;
done
```

trxcllnt:
@davebayer yes, that's the change I'd expect to see. Whether we want to make it an option or not should depend on how large the static lib is. If it's a few Mb, it's probably fine to keep it by default.

davebayer:
@trxcllnt it's like 2 MB, so I think it makes no difference :)

davebayer:
Thanks @trxcllnt!

PR #531, opened 2025-06-25
These python lib packages should depend on the C++ package

wence-:
This is cargo-culting the way other `libfoo` python packages depend on the C++ library.


features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
```diff
@@ -42,6 +42,7 @@ repos:
     - name: libucxx
       sub_dir: python/libucxx
       args: {install: *rapids_build_backend_args}
+      depends: [ucxx]
```
vyasr, line 45: This should probably also include ucxx-python.

vyasr (commented):
This is also what https://github.com/rapidsai/build-planning/issues/119 is for 😞

wence-:
> Yes, I think this is right. Thanks for fixing it.
> 
> Please bump this version to `25.8.6`:
> 
> https://github.com/rapidsai/devcontainers/blob/31fd259b050b3229b3fcb20515c49be9c1d63825/features/src/rapids-build-utils/devcontainer-feature.json#L4
> 
> That needs to be done every time you change something in `features/src/rapids-build-utils`, to ensure the updated code is released successfully.

Done

features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
wence-: Also done.

jameslamb:
Ah, I think I don't have write access here. Someone else like @trxcllnt or @bdice will have to merge this.

vyasr:
Done.

PR #532, opened 2025-06-30
Use CUDA 12.9 in devcontainers

jakirkham:
Now that RAPIDS is migrating from CUDA 12.8 to 12.9, update the devcontainers accordingly.

References:
* https://github.com/rapidsai/build-planning/issues/195
* https://github.com/rapidsai/build-planning/issues/173


matrix.yml
```diff
@@ -123,7 +123,7 @@ include:
   # FIXME 12.3, 12.4, 12.7 don't seem to have apt packages avaiable in the repos. Investigate if we end up really needing those.
   - { features: [*python, *dood, *gcc_13, { name: "cuda", version: "12.5", <<: *cccl_cuda_opts }, *clang_format_cccl, *clangd_dev, *cccl_dev], env: *gcc_env }
   - { features: [*python, *dood, *gcc_13, { name: "cuda", version: "12.6", <<: *cccl_cuda_opts }, *clang_format_cccl, *clangd_dev, *cccl_dev], env: *gcc_env }
-  - { features: [*python, *dood, *gcc_13, { name: "cuda", version: "12.8", <<: *cccl_cuda_opts }, *clang_format_cccl, *clangd_dev, *cccl_dev], env: *gcc_env }
+  - { features: [*python, *dood, *gcc_13, { name: "cuda", version: "12.9", <<: *cccl_cuda_opts }, *clang_format_cccl, *clangd_dev, *cccl_dev], env: *gcc_env }
```
trxcllnt, line 126: I think if we make the change on L5 and here, we will need to update the CCCL cuda-python CI jobs to 12.9 too.

jakirkham: Thanks Paul! 🙏

After looking upstream, it seems they already have largely moved to CUDA 12.9. Only the RAPIDS CI portion is still on CUDA 12.8. Given this, think it makes sense to update CCCL

Filed upstream issue:
* https://github.com/NVIDIA/cccl/issues/5103

Also submitted PR:
* https://github.com/NVIDIA/cccl/pull/5104

As it sounds like you have done this before, please take a look and let me know if more is needed over there. Also if there are more changes needed here, happy to update this PR too

jakirkham:
Thanks Paul! 🙏

Appreciate your help walking through this 🙂

PR #534, opened 2025-07-14
Update RAPIDS containers to Ubuntu 24.04

bdice:
Currently nightly builds fail for cuDF because of the use of `<format>`. This requires GCC 13 or newer.

The devcontainers are getting the OS default version of GCC (because RAPIDS devcontainers don't specify a version). The OS is Ubuntu 22.04 which ships with GCC 11. Updating to Ubuntu 24.04 should solve this problem. Alternatively we can pin the GCC version, but I like being able to use system-default toolchains where possible.

### DIFF
--- a/.devcontainer/cuda12.9-conda/devcontainer.json
+++ b/.devcontainer/cuda12.9-conda/devcontainer.json
@@ -5,7 +5,7 @@
     "args": {
       "CUDA": "12.9",
       "PYTHON_PACKAGE_MANAGER": "conda",
-      "BASE": "rapidsai/devcontainers:25.08-cpp-mambaforge-ubuntu24.04"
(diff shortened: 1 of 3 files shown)


trxcllnt:
Looks like there aren't any CUDA 12.0 Ubuntu24.04 apt repos, so probably gotta stick with 22.04 for those (if we still need them).

bdice:
The problem is coming from `xgboost`. `xgboost` depends on `librmm`, and the `cuml` development environment installs `xgboost` by default.

Do we need to make the devcontainers aware of xgboost, or add some special exclusion for librmm?

bdice:
Great. I filed a PR here with that change: https://github.com/rapidsai/cuml/pull/7006

trxcllnt (approved):

PR #535, opened 2025-07-15
Exclude xgboost packages

bdice:
Packages for `xgboost` depend on `librmm`. However, we don't want `librmm` to be pulled in as a conda dependency. One approach to fixing this is in https://github.com/rapidsai/cuml/pull/7006 but this might be a cleaner / more minimal solution.

### DIFF
--- a/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-conda-dependencies.sh
+++ b/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-conda-dependencies.sh
@@ -161,8 +161,10 @@ make_conda_dependencies() {
             done
         done
 
+        local xgboost_packages=(xgboost py-xgboost libxgboost);
+
         # shellcheck disable=SC2207
-        local conda_noinstall=($(rapids-python-pkg-names) $(rapids-python-conda-pkg-names));
+        local conda_noinstall=($(rapids-python-pkg-names) $(rapids-python-conda-pkg-names) ${xgboost_packages[@]});
 
         # Generate a combined conda env yaml file.
         conda-merge "${conda_env_yamls[@]}"                                                                                   \


trxcllnt:
I don't think we should hard-code this here. I think we should make xgboost an optional test dependency for cuML.

bdice:
I think this is a better fix than changing cuML, because cuML developers _do_ want to have xgboost installed in the developer conda environment for running the full test suite. Its scope of impact is narrower, too.

Even aside from the current usage in cuML, xgboost is the only non-RAPIDS package that depends on RAPIDS packages I can think of, and that's fundamentally breaking for the way our devcontainers work. There's no world in which we can have `xgboost` in the unified devcontainers environment unless we retool all our build scripts to support xgboost and add it to the devcontainers.

trxcllnt:
Then cuML should generate a `test-env.yml` to install build, runtime, and test dependencies (or alternatively, have a different "build_and_run" key we can use in this script, but that's probably a bigger ask).

trxcllnt:
Putting this here will exclude xgboost from being installed in any conda devcontainer, including the one in the non-combined one in the cuML repo. Not sure that's what we want.

jameslamb:
> _We reused the existing `all` key for devcontainers, but the primary purpose is for developer conda environments (not related to devcontainer usage)._

I think it could be useful to be able to set `dependencies.yaml` keys in the matrix, like this:

https://github.com/rapidsai/devcontainers/blob/0e927a99e207be250dfae06f1c5a7885898b9b70/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml#L177-L179

becoming this:

```yaml
 - name: cuml 
   path: cuml 
   git: {<<: *git_defaults, repo: cuml}
   dependency-file-generator-keys:
     - devcontainer
```

Then we could make devcontainers-specific choices project-by-project without affecting anything else about the projects.

But I'm not personally offering to implement that and not asking you to @bdice .

I've said what I need to here, I'm going to take myself out of this discussion. Merge whichever of these PRs (this or the `cuml` one) you want based on what you've heard in these discussions.

PR #536, opened 2025-07-16
Support repository dependency_keys in manifest.yaml

bdice:
This PR isolates a change from @trxcllnt in #479 so that we can unblock CI on this repository.

We need this so we can exclude `xgboost` from cuML's dependency list, since `libxgboost` brings in `librmm` conda packages, and that breaks devcontainers building everything from source.

Closes #535.

This temporarily breaks cuML devcontainers because it does not include the cuML test dependencies, but it unblocks cuML CI.

Planned follow-up actions:
- Create a cuML dependency file key called `devcontainers` with everything except `xgboost`
- Update the `manifest.yaml` to use that file key instead of the current set of keys (if the cuML PR merges first, we can modify this PR instead of needing a follow-up)

### DIFF
--- a/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-conda-dependencies.sh
+++ b/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-conda-dependencies.sh
@@ -45,7 +45,6 @@ make_conda_dependencies() {
 
     test ${#exclude[@]} -eq 0 && exclude=();
     test ${#include[@]} -eq 0 && include=();
-    test ${#key[@]} -eq 0 && key=(all);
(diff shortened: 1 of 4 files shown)


jameslamb (approved):
Thank you for considering this and to @trxcllnt for implementing it.

I agree, we need to get this onto `branch-25.08` and unblock CI.

jameslamb:
Every time you update one of the devcontainers features, you have to update the version to ensure it's published:

https://github.com/rapidsai/devcontainers/blob/0e927a99e207be250dfae06f1c5a7885898b9b70/features/src/rapids-build-utils/devcontainer-feature.json#L4

I just pushed https://github.com/rapidsai/devcontainers/pull/536/commits/c82f309868fb457e5ebb0414d2024adc882818cb doing that here.

PR #538, opened 2025-07-17
Add `cu++filt` package to windows devcontainer

davebayer:
In #530 I forgot to update the windows devcontainer, too. This PR adds the `cuxxfilt` package to windows devcontainer. The package is only couple of kBs


trxcllnt:
No, they're unrelated

PR #549, opened 2025-07-31
CCCL: Add gcc-15

alliepiper:
- Remove redundant images
- Add gcc15

_Marked as draft, ppas don't have packages yet_


trxcllnt:
I think this is addressed by https://github.com/rapidsai/devcontainers/pull/516, feel free to reopen if not.

PR #561, opened 2025-08-14
set glab cli git protocol to GIT_PROTOCOL

msarahan:
We currently can't use https protocol, which is hindering internal DLFW projects. I think this should fix things, and also brings things into parity with the github cli.


jameslamb (approved):
cool cool makes sense to me!

PR #611, opened 2025-11-11
Add libc++ to our clang devcontainers

miscco:
We also want to test clang-cuda against libc++ because that is a common use case from one of our customers


trxcllnt:
Need to update the devcontainer feature version

PR #628, opened 2025-12-04
Fix nvhpc modules

trxcllnt:


features/test/nvhpc/scenarios.json
```diff
@@ -0,0 +1,34 @@
...
+      }
+    }
+  },
+  "nvhpc_25_11": {
+    "image": "ubuntu:24.04",
+    "features": {
+      "nvhpc": {
+        "version": "25.11"
+      }
+    }
```
alliepiper: Does this framework support testing with the default/latest version, so we can catch these in the future?

PR #634, opened 2025-12-11
Add messages to identify which builds/tests are happening

bdice:
Sometimes when I run `build-all -j0` I have a hard time identifying which library is building (the name doesn't always appear clearly in the output). This adds a message prior to building or running tests to help clarify.


trxcllnt (changes requested):
Need to bump feature version, but other than that this is fine

bdice:
Bumped the version. This should be mergeable.

PR #636, opened 2025-12-14
Add CUDA 13.1 support

bdice:
## Summary
- Update `cuda_curr_max_rapids` to CUDA 13.1 in matrix.yml
- Add cuda13.1-conda and cuda13.1-pip devcontainer configurations
- Update build-all-rapids-repos.yml to use CUDA 13.1

Part of https://github.com/rapidsai/build-planning/issues/236

## Dependencies

This depends on all RAPIDS libraries having CUDA 13.1 in their `dependencies.yaml`.


jameslamb:
cuVS build errors here should hopefully be fixed by https://github.com/rapidsai/cuvs/pull/1686

```text
CMake Error in CMakeLists.txt:
  export called with target "cuvs_static" which requires target "CUTLASS"
  that is not in any export set.
```

([build link](https://github.com/rapidsai/devcontainers/actions/runs/20831321270/job/59845970825?pr=636#step:9:8170))

PR #637, opened 2025-12-16
Add cuOpt to unified devcontainers

bdice:
This adds cuOpt to the unified devcontainers. It releases on the same cadence as RAPIDS and has RAPIDS dependency infrastructure, so it's useful to be able to do test builds of cuOpt in a RAPIDS environments.

Depends on:
- https://github.com/NVIDIA/cuopt/pull/660
- https://github.com/NVIDIA/cuopt/pull/722
- https://github.com/NVIDIA/cuopt/pull/723
- https://github.com/NVIDIA/cuopt/pull/729


trxcllnt:
Looks like we need to add TBB to the pip devcontainers for cuOpt?

trxcllnt (approved):

trxcllnt (changes requested):
@bdice could you add cuopt to one of the matrix options [here](https://github.com/rapidsai/devcontainers/blob/main/.github/workflows/build-all-rapids-repos.yml#L42-L44)?

PR #692, opened 2026-04-30
Minimize cuOpt Boost dependencies

bdice:
## Summary
- keep only Boost headers, iostreams, and serialization needed by cuOpt/PaPILO
- skip `libboost-program-options-dev` -- this is optional and not needed for PaPILO with the options we provide (`PAPILO_NO_BINARIES=ON`)


bdice:
Merged to unblock follow-up work on cuOpt devcontainers (CI was passing and these changes are similar to the changes I already landed upstream in cuOpt).

PR #693, opened 2026-04-30
Build and publish devcontainers to `ghcr.io` and use `cacheFrom` to speed up devcontainer launch times

trxcllnt:
Depends on https://github.com/rapidsai/shared-workflows/pull/530

Example of the container images in `ghcr.io`: https://github.com/trxcllnt/devcontainers/pkgs/container/devcontainers%2Fdevcontainer


.github/actions/setup-runner-env/action.yml
```diff
@@ -26,28 +26,28 @@ runs:
 
     - if: runner.environment == 'self-hosted'
       name: Setup Node.js
-      uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+      uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e #v6
```
bdice: For consistency.
```suggestion
      uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
```

.github/workflows/build-devcontainers.yml
```diff
@@ -0,0 +1,57 @@
...
+        default: '["conda", "pip"]'
+      push:
+        type: string
+        default: true
+
+jobs:
+  build:
+    name: Build devcontainers
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/build-devcontainers.yaml@fea/build-devcontainer
```
bdice: Maybe needs to be updated after that merges? Same for other references.

trxcllnt: yes

.github/actions/setup-runner-env/action.yml
trxcllnt:
```suggestion
      uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
```

PR #694, opened 2026-05-06
ensure latest tag is used to determine versions

jameslamb:
Contributes to https://github.com/rapidsai/github-infrastructure/issues/63

Mistakes in forward-mergers can sometimes lead to a weird mix of tags in `git` history, which can trip up logic that relies on those tags to compute things like versions for image URIs.

`git describe --tags --first-parent` ensures we get the expected behavior (version = tag "closest" to HEAD) even when mistakes are made in the forward-merger process.

See https://github.com/rapidsai/docker/pull/867 and https://github.com/rapidsai/github-infrastructure/issues/63 for much more detail.

### How I tested this

These commands are equivalent on `main` here.

```console
$ git checkout main
$ git fetch upstream --tags
$ git describe --abbrev=0 --tags
v26.06.00a

$ git describe --abbrev=0 --tags --first-parent
v26.06.00a
```

Just doing this preventatively.

### DIFF
--- a/.github/actions/devcontainer-json/action.sh
+++ b/.github/actions/devcontainer-json/action.sh
@@ -7,7 +7,7 @@ os="${1:-"ubuntu:22.04"}";
 features="${2:-"[]"}";
 container_env="${3:-"null"}";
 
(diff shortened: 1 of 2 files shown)


gforsyth (approved):
Big fan of preventative measure.

:shipit:

trxcllnt (approved):

PR #696, opened 2026-05-11
Replace CUDA 13.1 with 13.2 for RAPIDS devcontainers

bdice:
## Summary
- Update `cuda_curr_max_rapids` to CUDA 13.2 in matrix.yml
- Rename cuda13.1-conda and cuda13.1-pip devcontainer configurations to cuda13.2
- Update build-all-rapids-repos.yml to use CUDA 13.2

Part of https://github.com/rapidsai/build-planning/issues/265

## Dependencies

This depends on all RAPIDS libraries having CUDA 13.2 in their `dependencies.yaml`.


jameslamb (approved):

trxcllnt (approved):

bdice:
This is broken because things are being cloned from `main` instead of the target branch `release/26.06`...

wence-:
I think if someone kicks the builds this should now work again, we were missing tags in various places?

bdice:
This is currently blocked by #701. Or, at least, if we care about green CI...

This repo's CI currently doesn't work on release branches, and I'd like to fix that before bumping the version here.

PR #700, opened 2026-05-15
Update tags for 26.08

wence-:


gforsyth (approved):
Admin-merging this to get the 26.08 devcontainers built

PR #704, opened 2026-05-19
Use mamba to create devcontainer conda envs

bdice:
## Description
Switch rapids-make-conda-env from `conda env create --solver=libmamba` to `mamba env create`.

This keeps environment creation on mamba directly instead of invoking conda's env frontend with the libmamba solver.

Error outputs from mamba tend to be clearer, and it doesn't spam logs with a spinner like `- / | \ - / | \`...

### DIFF
--- a/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-conda-env.sh
+++ b/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-conda-env.sh
@@ -54,7 +54,7 @@ make_conda_env() {
             cat "${new_env_path}";
             echo "";
 
-            conda env create "${q[@]}" -n "${env_name}" -f "${new_env_path}" --solver=libmamba;
+            mamba env create "${q[@]}" -n "${env_name}" -f "${new_env_path}";
         # If the conda env does exist but it's different from the generated one,
         # print the diff between the envs and update it
         elif ! diff -BNqw "${old_env_path}" "${new_env_path}" >/dev/null 2>&1; then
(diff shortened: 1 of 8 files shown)


features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-conda-env.sh
vyasr, line 1: It looks like as of https://github.com/rapidsai/devcontainers/pull/249 we are always making conda envs from scratch rather than updating. That being the case, can we narrow what's done conditionally here and merge more of the code paths for creating/updating? The comments are also a bit out of date relative to the current state of how it's working now.

vyasr (changes requested):

features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-conda-env.sh
trxcllnt: Yeah we always delete and recreate the env instead of updating it, because conda doesn't clean up after itself.

trxcllnt (changes requested):
Could you also bump the versions in the respective features' `devcontainer-feature.json` file?

trxcllnt:
Does mamba not respect the `CONDA_ALWAYS_YES=true` envvar? Or does it need to be `MAMBA_ALWAYS_YES=true`?

trxcllnt:
Ignoring `CONDA_PKGS_DIRS` may be a [mamba bug](https://github.com/mamba-org/mamba/issues/4250)?

trxcllnt:
fixed mamba https://github.com/rapidsai/devcontainers/actions/runs/26188377829/job/77049984252?pr=704#step:9:3069

PR #705, opened 2026-05-19
Switch from building RMM to building CCCL in UCXX

pentschev:
CCCL support for UCXX was added in https://github.com/rapidsai/ucxx/pull/641, and https://github.com/rapidsai/ucxx/pull/656 is removing RMM support. Switching to CCCL here is necessary.

### DIFF
--- a/features/src/rapids-build-utils/devcontainer-feature.json
+++ b/features/src/rapids-build-utils/devcontainer-feature.json
@@ -1,7 +1,7 @@
 {
   "name": "NVIDIA RAPIDS devcontainer build utilities",
   "id": "rapids-build-utils",
-  "version": "26.8.0",
+  "version": "26.8.1",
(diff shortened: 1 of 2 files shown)


vyasr (approved):

trxcllnt (changes requested):
Also need to bump the patch version in [this file](https://github.com/rapidsai/devcontainers/blob/d5f1538e357d8da11dc98b89a76699614ba2bbfa/features/src/rapids-build-utils/devcontainer-feature.json#L4)

pentschev:
> Also need to bump the patch version in [this file](https://github.com/rapidsai/devcontainers/blob/d5f1538e357d8da11dc98b89a76699614ba2bbfa/features/src/rapids-build-utils/devcontainer-feature.json#L4)

Done.

trxcllnt (approved):

PR #707, opened 2026-05-20
Use `-DUCXX_ENABLE_RMM=ON` because it's necessary to configure `ucxx::python`

trxcllnt:


pentschev (commented):
https://github.com/rapidsai/ucxx/pull/656 should remove that requirement. Can we wait for that to be merged perhaps?

trxcllnt:
@pentschev Ok, I'll close this

PR #708, opened 2026-05-21
Add nvforest to cuML devcontainer dependencies

csadorf:
Add `nvforest` to the cuML C++ devcontainer dependency list so cuML builds after nvForest when using the RAPIDS build-utils manifest.

Follow-up to https://github.com/rapidsai/cuml/pull/8048

### DIFF
--- a/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
+++ b/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
@@ -194,7 +194,7 @@ repos:
     cpp:
       - name: cuml
         sub_dir: cpp
-        depends: [rmm, raft, cuvs]
+        depends: [rmm, raft, cuvs, nvforest]
         parallelism:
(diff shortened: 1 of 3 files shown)


bdice (approved):

features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
```diff
@@ -194,7 +194,7 @@ repos:
     cpp:
       - name: cuml
         sub_dir: cpp
-        depends: [rmm, raft, cuvs]
+        depends: [rmm, raft, cuvs, nvforest]
```
trxcllnt: The list of repos is sorted, so we'll need to move the ` - name: nvforest` block above the cuml entry here. Also need to update the version in this feature's `devcontainer-feature.json` file.

trxcllnt (changes requested):
see comment

trxcllnt (approved):

PR #709, opened 2026-05-22
Use the new `SCCACHE_PID_FILE` envvar so the sccache daemon doesn't randomly change pids

trxcllnt:


bdice (approved):
Looks fine to me -- I'll report back if I see further issues!

PR #691, opened 2026-04-24
fix: resolve template-injection issues and harden workflows

gforsyth:
- **fix(ci): pin all third-party actions**
- **fix: replace all template injection sites with env vars**
- **fix: suppress GITHUB_ENV warning (non-dangerous)**
- **fix: ignore dangerous-trigger warning**
- **fix: grant permissions explicitly per-job**
- **fix: suppress secrets-inherit warnings**
- **fix: disable `persist-credentials` to prevent credential leakage**
- **fix: remediate or ignore all template-injection warnings**

PR #714, opened 2026-06-01
fix(ci): fix configuration for breaking change notification workflow

jameslamb:
Contributes to https://github.com/rapidsai/shared-workflows/issues/560

Fixes misconfigurations in the GitHub Actions workflow that generates Notifications when PRs are labeled `breaking`.

Any other changes come from making the configuration for this workflow identical across all RAPIDS repos.

### DIFF
--- a/.github/zizmor.yml
+++ b/.github/zizmor.yml
@@ -1,0 +1,9 @@
+rules:
+  unpinned-uses:
+    config:
+      policies:
+        # We require SHA-pinning for all workflows and actions _except_ for those from
+        # rapidsai/shared-workflows and rapidsai/shared-actions
+        "rapidsai/shared-workflows/*": any
+        "rapidsai/shared-actions/*": any
+        "*": hash-pin
--- a/.github/workflows/trigger-breaking-change-alert.yaml
+++ b/.github/workflows/trigger-breaking-change-alert.yaml
@@ -1,22 +1,26 @@
 name: Trigger Breaking Change Notifications
 
-on:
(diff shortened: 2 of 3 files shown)


.github/zizmor.yml
```diff
@@ -0,0 +1,9 @@
+rules:
```
jameslamb, line 1: `zizmor` was failing on this branch, like this:

```text
error[unpinned-uses]: unpinned action reference
  --> .github/workflows/trigger-breaking-change-alert.yaml:21:11
   |
21 |     uses: rapidsai/shared-workflows/.github/workflows/breaking-change-alert.yaml@main
   |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ action is not pinned to a hash (required by blanket policy)
   |
   = note: audit confidence → High

76 findings (28 ignored, 47 suppressed): 0 informational, 0 low, 0 medium, 1 high
```

This fixes that, in the way we have across the rest of RAPIDS.

jakirkham: Thanks for adding this James! 🙏

Yeah zizmor was added to this repo in PR: https://github.com/rapidsai/devcontainers/pull/691

That work preceded the more recent in issue: https://github.com/rapidsai/build-planning/issues/275

So it makes sense that it was missing some bits that got standardized later

jakirkham (approved):
Thanks James! 🙏

PR #715, opened 2026-06-04
Add new cudf-streaming C++ and Python builds

vyasr:
These new libraries contain the pieces of rapidsmpf that rely on cudf. As a result, the dependency ordering is also now switched because rapidsmpf will no longer rely on cudf.

### DIFF
--- a/features/src/rapids-build-utils/devcontainer-feature.json
+++ b/features/src/rapids-build-utils/devcontainer-feature.json
@@ -1,7 +1,7 @@
 {
   "name": "NVIDIA RAPIDS devcontainer build utilities",
   "id": "rapids-build-utils",
-  "version": "26.8.3",
+  "version": "26.8.4",
   "description": "A feature to install the RAPIDS devcontainer build utilities",
   "containerEnv": {
     "BASH_ENV": "/etc/bash.bash_env"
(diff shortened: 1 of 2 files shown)


trxcllnt (approved):

bdice (approved):

trxcllnt:
It looks like rapidsmpf still [depends on](https://github.com/rapidsai/devcontainers/actions/runs/26982605720/job/79625156717?pr=715#step:9:12409) pylibcudf in Cython code. Is that right?

vyasr:
Yes, it's a circular dependency so we're going to have to break out of some catch 22s to get everything merged in without breaking too many things at once.  rapidsmpf development is currently frozen, so I think our best option is to merge this PR now and break rapidsmpf so that the the cudf builds on https://github.com/rapidsai/cudf/pull/22786 pass, then merge that cudf PR, which will allow cudf-polars to start working properly without the parts of rapidsmpf that induce the circular dependency. Then we can open a PR to rapidsmpf that removes the problematic components, unfreeze main, and merge that branch, which should resolve everything.

trxcllnt:
@vyasr can we go back to building rapidsmpf after cudf? or does `cudf_streaming` depend on `librapidsmpf` being built first?

vyasr:
`libcudf_streaming` (the C++ library) depends on `librapidsmpf`. Is there a reason that you want to build cudf first?

PR #716, opened 2026-06-05
Support git worktrees in RAPIDS script generation

bdice:
## Summary
- Detect checked-out RAPIDS repos with `git -C <path> rev-parse --is-inside-work-tree` instead of testing for a `.git` directory
- Allow RAPIDS build utility scripts to recognize git worktrees, where `.git` is a file instead of a directory
- Update the shared clone helper so local worktree origins and existing worktree targets are treated as git repositories
- Bump `rapids-build-utils` from `26.8.4` to `26.8.5` and `utils` from `26.8.1` to `26.8.2`

## Motivation
This is necessary to support devcontainers that use git worktrees. Worktree checkouts store `.git` as a file pointing at the real git metadata, so the previous `-d <repo>/.git` checks treated valid worktree repos as missing. In `rapids-generate-scripts`, that meant only aggregate/no-op scripts were generated.

## Testing
- `bash -n` passed for the 10 modified shell scripts
- `git diff --check` passed
- confirmed there are no remaining `.git` directory predicates in the repo
- created a temporary git worktree inside the devcontainer and verified the updated clone helper accepts it as a local git repository
- patched the installed script in an RMM worktree devcontainer and verified `rapids-generate-scripts --repo rmm` generated `build-rmm` and `test-rmm-python`
- in the RMM worktree devcontainer, `build-all -j0` passed
- in the RMM worktree devcontainer, `test-all-python` passed: `1780 passed, 6 skipped`

### DIFF
--- a/features/src/rapids-build-utils/devcontainer-feature.json
+++ b/features/src/rapids-build-utils/devcontainer-feature.json
@@ -1,7 +1,7 @@
 {
   "name": "NVIDIA RAPIDS devcontainer build utilities",
   "id": "rapids-build-utils",
-  "version": "26.8.4",
+  "version": "26.8.5",
   "description": "A feature to install the RAPIDS devcontainer build utilities",
   "containerEnv": {
     "BASH_ENV": "/etc/bash.bash_env"
--- a/features/src/rapids-build-utils/opt/rapids-build-utils/bin/checkout-same-branch.sh
+++ b/features/src/rapids-build-utils/opt/rapids-build-utils/bin/checkout-same-branch.sh
@@ -17,6 +17,11 @@
 # shellcheck disable=SC1091
 . rapids-generate-docstring;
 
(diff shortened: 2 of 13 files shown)


vyasr (commented):
Seems robust enough for the devcontainers use case.

vyasr (approved):

trxcllnt (changes requested):
Need to bump the feature version, otherwise looks fine to me. Are there any other places that check for `-d .git` that also need to be updated?

bdice:
@trxcllnt This should be ready for another look. I'm not sure if it makes sense to consolidate the `repo_exists` somewhere.

trxcllnt:
Yeah it'd be nice to make a little `local-repo-clone-exists` script or something, but it's not critical.

trxcllnt (approved):

PR #725, opened 2026-06-25
Bump clang-tools version for CCCL

Jacobfaib:


Jacobfaib:
Hmmm seems like ubuntu 20.04 is in a strange state of support limbo. It's technically EOL according to ubuntu but LLVM still kinda supports it? All of the arm64 binaries are missing for clang-tools there as a result

https://github.com/llvm/llvm-project/issues/187399

PR #731, opened 2026-07-06
Add MPI to some dev containers

Jacobfaib:


trxcllnt (changes requested):
This PR will stop publishing the images for the modified lines under the current tag, and start publishing new ones with `-openmpi5.0.10` in the tag instead. Is that intended, or do you want to keep the current images and add new ones that contain openmpi?

Jacobfaib:
Original intent was to keep the same images but include openmpi in them transparently. But now we are considering not adding openmpi at all, probably will close the PR anyways.

PR #732, opened 2026-07-07
Update CCCL clang-format to 22

bernhardmgruber:


Jacobfaib:
Note, I tried this previously https://github.com/rapidsai/devcontainers/pull/725 but there are some issues with LLVM on old Ubuntu (20.04) and ARM specifically

bdice (approved):

bernhardmgruber:
Why do we still need Ubuntu 20.4 images? Can we drop them?

trxcllnt:
No, Ubuntu 20.04 is the only OS release supported by CTK 12.0/12.9 and these specific host compiler versions.

bernhardmgruber:
I asked `claude` to help me here, and apparently there are no LLVM 22 builds for Ubuntu 20.04. It suggested to stay at LLVM 21 for the affected Ubuntu versions, and I think this may be a good compromise.

trxcllnt (approved):

PR #736, opened 2026-07-08
fix: propagate dependency generation failures

bdice:
## Summary

- stage each dependency-file-generator result and wait for every background job
- preserve generator stderr and return nonzero without emitting combined output when any job fails
- skip conventional dependency keys that are absent from a repository config while deduplicating repeated keys
- cover the behavior for both pip requirements and conda environments
- bump `rapids-build-utils` to `26.8.8`

## Context

In rapidsai/kvikio#996, `rapids-dependency-file-generator` wrote a partial requirements list before rejecting an invalid pip map. The devcontainers wrapper discarded stderr and collected successful process-substitution output, so the generator failure was hidden and the incomplete requirements file was installed. The later build then surfaced the unrelated-looking missing-Cython error.

This change makes the wrapper fail at the dependency-generation boundary and keeps partial child output out of its stdout. The companion generator-side change is rapidsai/dependency-file-generator#188.

## Testing

- `devcontainer features test --project-folder ./features --features rapids-build-utils --base-image ubuntu:22.04`
- `bash -n` on both modified dependency scripts and the feature test
- `git diff --check`

### DIFF
--- a/features/src/rapids-build-utils/devcontainer-feature.json
+++ b/features/src/rapids-build-utils/devcontainer-feature.json
@@ -1,7 +1,7 @@
 {
   "name": "NVIDIA RAPIDS devcontainer build utilities",
   "id": "rapids-build-utils",
-  "version": "26.8.7",
+  "version": "26.8.8",
   "description": "A feature to install the RAPIDS devcontainer build utilities",
   "containerEnv": {
     "BASH_ENV": "/etc/bash.bash_env"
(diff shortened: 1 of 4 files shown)


trxcllnt:
IIRC we don't return the error exit code because we don't want to fail when generating for additional file keys that some projects have and others don't. We do this in the DLFW builds when we pass `--key py_build --key py_test_cudf` etc.

PR #741, opened 2026-07-15
Install debugger init files

Jacobfaib:
Install global `.gdbinit` and `.lldbinit` files for CCCL devcontainers

### DIFF
--- a/features/src/cccl-dev/install.sh
+++ b/features/src/cccl-dev/install.sh
@@ -86,6 +86,12 @@ append_to_all_bashrcs "$(cat .bashrc | envsubst '$LIT_VERSION $DOXYGEN_VERSION $
 # export envvars in /etc/profile.d
 add_etc_profile_d_script cccl-dev "$(cat .bashrc | envsubst '$LIT_VERSION $DOXYGEN_VERSION $USERHOME')";
 
+cccl_dir="${USERHOME}/cccl"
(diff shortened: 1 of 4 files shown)


Jacobfaib:
I did notice that on some devcontainers, the debuggers are not installed by default (at least according to our CI). Can we install them as well (depending on installed compilers of course)?

features/src/cccl-dev/install.sh
```diff
@@ -86,6 +86,12 @@ append_to_all_bashrcs "$(cat .bashrc | envsubst '$LIT_VERSION $DOXYGEN_VERSION $
 # export envvars in /etc/profile.d
 add_etc_profile_d_script cccl-dev "$(cat .bashrc | envsubst '$LIT_VERSION $DOXYGEN_VERSION $USERHOME')";
 
+cccl_dir="${USERHOME}/cccl"
+# install global lldbinit and gdbinit, append in case there is a pre-existing one
+# installed by the OS or packages
+cat .lldbinit | sed -e "s%@CCCL_DIR@%${cccl_dir}%" >> "${USERHOME}/.lldbinit";
```
Jacobfaib, line 92: Also yes, currently this `sed` doesn't do anything, but if it ever does then we don't have to change this

trxcllnt:
LGTM, just need to bump the patch version in `features/src/cccl-dev/devcontainer-feature.json`

trxcllnt (approved):

PR #751, opened 2026-07-29
Specify `use_cuda_wheels`

KyleFromNVIDIA:
In order to support https://github.com/rapidsai/pre-commit-hooks/issues/132, matrix parameters like `use_cuda_wheels` have to be explicit so that the `verify-dependencies` hook doesn't produce false positives. Make it explicit so we can use the hook.

### DIFF
--- a/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
+++ b/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
@@ -5,7 +5,7 @@ x-git-defaults: &git_defaults
 x-rapids-build-backend-args: &rapids_build_backend_args |
   --config-settings "skbuild.strict-config=false"
   --config-settings "rapidsai.disable-cuda=true"
-  --config-settings "rapidsai.matrix-entry=cuda=${CUDA_VERSION_MAJOR_MINOR};use_cuda_wheels=true;cuda_suffixed=false"
+  --config-settings "rapidsai.matrix-entry=cuda=${CUDA_VERSION_MAJOR_MINOR};use_cuda_wheels=false"
 repos:
   - name: rmm
     path: rmm
--- a/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
+++ b/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
@@ -5,7 +5,7 @@ x-git-defaults: &git_defaults
 x-rapids-build-backend-args: &rapids_build_backend_args |
   --config-settings "skbuild.strict-config=false"
   --config-settings "rapidsai.disable-cuda=true"
-  --config-settings "rapidsai.matrix-entry=cuda=${CUDA_VERSION_MAJOR_MINOR}"
+  --config-settings "rapidsai.matrix-entry=cuda=${CUDA_VERSION_MAJOR_MINOR};use_cuda_wheels=true;cuda_suffixed=false"
 repos:
   - name: rmm
     path: rmm
--- a/features/src/rapids-build-utils/devcontainer-feature.json
+++ b/features/src/rapids-build-utils/devcontainer-feature.json
@@ -1,7 +1,7 @@
 {
   "name": "NVIDIA RAPIDS devcontainer build utilities",
   "id": "rapids-build-utils",
-  "version": "26.10.0",
+  "version": "26.10.1",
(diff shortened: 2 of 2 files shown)


features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
```diff
@@ -5,7 +5,7 @@ x-git-defaults: &git_defaults
 x-rapids-build-backend-args: &rapids_build_backend_args |
   --config-settings "skbuild.strict-config=false"
   --config-settings "rapidsai.disable-cuda=true"
-  --config-settings "rapidsai.matrix-entry=cuda=${CUDA_VERSION_MAJOR_MINOR}"
+  --config-settings "rapidsai.matrix-entry=cuda=${CUDA_VERSION_MAJOR_MINOR};use_cuda_wheels=true;cuda_suffixed=false"
```
jameslamb: I think we pretty intentionally do NOT want to use CUDA wheels in devcontainers, because we rely on a system-installed CTK instead.

https://github.com/rapidsai/devcontainers/blob/1c973ffa0a768547e1cab6f622426ef8ac4b6539/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-pip-dependencies.sh#L173-L179

jameslamb: And Python libraries are build with `--editable --no-deps`, so the value of `cuda_suffixed` should (I think) not matter:

https://github.com/rapidsai/devcontainers/blob/1c973ffa0a768547e1cab6f622426ef8ac4b6539/features/src/rapids-build-utils/opt/rapids-build-utils/bin/tmpl/python.build.editable.tmpl.sh#L18

jameslamb (changes requested):
I don't think either of these changes match with how the devcontainers are used today in RAPIDS. See my inline comments.

features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
```diff
@@ -5,7 +5,7 @@ x-git-defaults: &git_defaults
 x-rapids-build-backend-args: &rapids_build_backend_args |
   --config-settings "skbuild.strict-config=false"
   --config-settings "rapidsai.disable-cuda=true"
-  --config-settings "rapidsai.matrix-entry=cuda=${CUDA_VERSION_MAJOR_MINOR}"
+  --config-settings "rapidsai.matrix-entry=cuda=${CUDA_VERSION_MAJOR_MINOR};use_cuda_wheels=false"
```
jameslamb, line 8: The practical effects of this will be to exclude CUDA wheels (e.g. `cuda-toolkit`, `nvidia-nvcomp-cu13`) from build-time and runtime dependencies.

Should be fine for commands like `build-{library}-python-editable`, because there packages are built with `--no-build-isolation` anyway, so already do not care about dependencies in the `[build-system]` table of the wheel.

https://github.com/rapidsai/devcontainers/blob/1c973ffa0a768547e1cab6f622426ef8ac4b6539/features/src/rapids-build-utils/opt/rapids-build-utils/bin/tmpl/python.build.editable.tmpl.sh#L17-L19

I'm unsure how it'll affect `install-{library}-python` or `build-{library}-python-wheel`. Have you tested those?

jameslamb (approved):
Every time you update a devcontainer feature, you need to also bump it's version so the changes will be published. Please bump this to `26.10.1`:

https://github.com/rapidsai/devcontainers/blob/1c973ffa0a768547e1cab6f622426ef8ac4b6539/features/src/rapids-build-utils/devcontainer-feature.json#L4

Otherwise, I'm marking this "approve" in the sense "ok with me if you want to try it". I'm not certain what the impact would be on `install-{library}-python` or `build-{library}-python-wheel`, but if some mix of CI and manual testing makes you confident in this change, then great.

Also please do add a short description answering the question "why do this?", linking to https://github.com/rapidsai/pre-commit-hooks/pull/147

features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
KyleFromNVIDIA: I tested locally and confirmed that `install-libcudf-python` now works where it was failing before.

trxcllnt (approved):

PR #753, opened 2026-07-31
Specify `cuda_suffixed=false`

KyleFromNVIDIA:
The fallback entries in `dependencies.yaml` previously had an implicit `cuda_suffixed: "false"` that we didn't need to specify here. Since it is now explicit, we have to specify it here too.

### DIFF
--- a/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
+++ b/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
@@ -5,7 +5,7 @@ x-git-defaults: &git_defaults
 x-rapids-build-backend-args: &rapids_build_backend_args |
   --config-settings "skbuild.strict-config=false"
   --config-settings "rapidsai.disable-cuda=true"
-  --config-settings "rapidsai.matrix-entry=cuda=${CUDA_VERSION_MAJOR_MINOR};use_cuda_wheels=false"
+  --config-settings "rapidsai.matrix-entry=cuda=${CUDA_VERSION_MAJOR_MINOR};use_cuda_wheels=false;cuda_suffixed=false"
 repos:
   - name: rmm
     path: rmm
--- a/features/src/rapids-build-utils/devcontainer-feature.json
+++ b/features/src/rapids-build-utils/devcontainer-feature.json
@@ -1,7 +1,7 @@
 {
   "name": "NVIDIA RAPIDS devcontainer build utilities",
   "id": "rapids-build-utils",
-  "version": "26.10.1",
+  "version": "26.10.2",
   "description": "A feature to install the RAPIDS devcontainer build utilities",
   "containerEnv": {
     "BASH_ENV": "/etc/bash.bash_env"


trxcllnt (approved):

features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
```diff
@@ -5,7 +5,7 @@ x-git-defaults: &git_defaults
 x-rapids-build-backend-args: &rapids_build_backend_args |
   --config-settings "skbuild.strict-config=false"
   --config-settings "rapidsai.disable-cuda=true"
-  --config-settings "rapidsai.matrix-entry=cuda=${CUDA_VERSION_MAJOR_MINOR};use_cuda_wheels=false"
+  --config-settings "rapidsai.matrix-entry=cuda=${CUDA_VERSION_MAJOR_MINOR};use_cuda_wheels=false;cuda_suffixed=false"
```
jameslamb, line 8: 1. reminder that you need to bump the version every time you change a feature: https://github.com/rapidsai/devcontainers/blob/77ff0aaca1bd0c88fb68bc1d16d7f1b1d9a9bd8c/features/src/rapids-build-utils/devcontainer-feature.json#L4
2. I get how this could work for the all-of-the-projects devcontainers, but how will it work for repo-specific ones? e.g. in the `cuml` pip devcontainer, don't we need to install the `libcuvs-cu12` wheel (suffixed) at build time, since we're not rebuilding cuVS from source there?

jameslamb (commented):
Looks like Paul already approved and he understands this stuff better than me, so leaving a non-blocking review. But I do have some comments.

features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
KyleFromNVIDIA: How did it work before? We would have had to specify `cuda_suffixed: "true"`, and we were not. This just makes explicit what was already implicitly there.

jameslamb: Fair enough, I don't know.

trxcllnt: This PR edits the `--config-settings` that gets passed to `pip install`, but that doesn't affect the dependencies installed by `rapids-make-pip-env` or `rapids-make-conda-env`.

The devcontainers install all their dependencies by running `rapids-dependency-file-generator` on all the repos' `dependencies.yaml` first, creates a huge conda `env.yml` or `requirements.txt`, and installs whatever is in those.

Does this PR need to edit either of these files instead?
* https://github.com/rapidsai/devcontainers/blob/77ff0aaca1bd0c88fb68bc1d16d7f1b1d9a9bd8c/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-pip-dependencies.sh#L162-L179
* https://github.com/rapidsai/devcontainers/blob/77ff0aaca1bd0c88fb68bc1d16d7f1b1d9a9bd8c/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-conda-dependencies.sh#L158-L162

trxcllnt (changes requested):
needs to bump the feature version (thanks @jameslamb!)

features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
KyleFromNVIDIA: I already confirmed locally that this change fixes the issues I was having with https://github.com/rapidsai/cudf/pull/23472 after making `cuda_suffixed: "false"` explicit.

KyleFromNVIDIA: `pip install` is exactly the problem. It's trying to get the dependency list from the freshly-built package, and RBB+DFG says "no matrix entry with that combo". This fixes that. (True, we're not actually using this dependency list, but RBB still has to provide it, and this makes it possible.)

trxcllnt: I think I follow. `cudf/dependencies.yaml` has been updated to explicitly require `cuda_suffixed=true|false` to be defined.

Even though we are not installing dependencies, `pip install --no-build-isolation --no-deps --editable cudf/python` is still running `rapids-dependency-file-generator` and needs `cuda_suffixed` to be in the `--matrix` argument, otherwise it will fail with the error in the cudf PR.

trxcllnt (approved):

features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml
KyleFromNVIDIA: Yes, that is correct.

PR #754, opened 2026-08-05
Require DFG >=1.22

KyleFromNVIDIA:
In order for projects to use the new `constraints` output type in v1.22.0, we have to ensure that rapids-dependency-file-generator is always v1.22.0 or greater.

Contributes to https://github.com/rapidsai/build-planning/issues/311


jameslamb (approved):
great, thank you

PR #756, opened 2026-08-18
Add UCX version up to 1.21.0

jakirkham:
Part of issue: https://github.com/rapidsai/build-planning/issues/314

Backfill old UCX versions all the way up to UCX 1.21.0.


jakirkham:
Also please let us know if we need to [add something to this directory]( https://github.com/rapidsai/devcontainers/tree/main/features/test/ucx ) and if so what to add

trxcllnt (approved):
LGTM 👍

jameslamb (approved):

jakirkham:
Thank you Paul and James! 🙏

One of the CI jobs looked stuck during the build. So have restarted it. Let's see how that goes

trxcllnt:
I think cugraph just has a few objects that take 45 minutes or so to compile

PR #757, opened 2026-08-19
Set cuda_suffixed=false;use_cuda_wheels=false in devcontainers dependency generation

bdice:
We are adopting a new convention where dependency lists explicitly require `cuda_suffixed` and `use_cuda_wheels` to be set. Currently cudf devcontainers fail on the `depends_on_kvikio` list when only the CUDA version is set and the suffix/wheels settings are left unset. This sets those values explicitly to help `rapids-dependency-file-generator` do the right thing.


trxcllnt:
@KyleFromNVIDIA did this in #751 and #753

bdice:
Closing as superseded by #751 and #753. The remaining issue is that the cuDF devcontainer lock resolves rapids-build-utils 26.10 to the older 26.10.0 feature.

bdice:
I think I launched my container wrong somehow. I got version `26.10.0` of rapids-build-utils instead of the latest. I think it's something stale on my end.

PR #761, opened 2026-09-01
Use default dependency key for RapidsMPF

bdice:
Remove RapidsMPF's `dependency_keys` override so `rapids-build-utils` uses the default `all` key.

`ray-default>=2.55.1` now resolves with PyTorch on both `linux-64` and `linux-aarch64`, so the workaround introduced for rapidsai/rapidsmpf#906 is no longer needed.


trxcllnt (changes requested):
Needs feature version bump, otherwise looks good

trxcllnt (approved):

PR #766, opened 2026-09-09
Fix unified pip devcontainer UCX base tags

msarahan:
**Posted by Codex (gpt-5.6-sol) on Michael Sarahan's behalf; treat this message as LLM-generated.**

The `release/26.10` image matrix builds UCX 1.21 (as of NVIDIA/devcontainers#756), but its checked-in unified pip devcontainers still requested UCX 1.19 base images. Those tags cannot be produced by the current branch matrix, so this PR updates the CUDA 12.9 and 13.3 pip bases to UCX 1.21.

The first full matrix exposed a second UCX compatibility problem hidden by the old configuration. The PyPI OpenMPI wheel bundles UCX 1.20 libraries. Installing that wheel while RAPIDS explicitly selects system UCX 1.21 creates a mixed runtime and leaves new `libucp` symbols unresolved during C++ links. When system UCX is preferred, the pip environment now excludes and removes the PyPI OpenMPI wheel, including from reused virtual environments. The base image already supplies OpenMPI built against its system UCX, so MPI and UCX remain one consistent stack.

This PR deliberately does not change `CUDA_HOME` to address the separate cuTile failures seen in the first run. RAPIDS build-planning issue https://github.com/rapidsai/build-planning/issues/324 establishes the intended dependency boundary: projects declare the `cuda-tile` Python build dependency, while `tileiras` comes from the system CUDA Toolkit selected for the build. The corresponding work is in rapidsai/cugraph#5646 and NVIDIA/cuvs#2564. Redirecting the entire conda devcontainer to a package-managed toolkit would hide that dependency problem and change the toolkit seen by unrelated CUDA consumers.

Related dependency work:

- https://github.com/rapidsai/build-planning/issues/324
- https://github.com/rapidsai/cugraph/pull/5646
- https://github.com/NVIDIA/cuvs/pull/2564

PR #767, opened 2026-09-09
Prevent release image bootstrap mismatches

msarahan:
**Posted by Codex (GPT-5) on Michael Sarahan's behalf. Treat this message as LLM-generated.**

## Why

The 26.12 version update triggered the release workflow before `v26.12.00a` existed. Image producers derive their namespace from the most recent reachable tag, so they published successful `26.10-*` images while the newly updated unified devcontainers requested `26.12-*` images. The alpha tag appeared nine minutes after the workflow began. A later retry of failed jobs could not recover because GitHub did not rerun the already-successful producer jobs.

This creates an avoidable bootstrap dependency: the version-update workflow needs to publish the new namespace before the tag that currently defines that namespace can be created. The committed `VERSION` file is already updated atomically with the consumer definitions and is therefore the appropriate source of truth for producer tags.

The incident also exposed a second class of mismatch: `matrix.yml` can change an image component version without the checked-in unified devcontainers changing their `BASE` references. Without validation, that drift is only discovered after all producer builds complete and consumers attempt to pull an image that will never be published.

PR #766 applies the UCX alignment to `release/26.10` for forward-merging. This branch currently contains the equivalent main-branch alignment; that portion will disappear from this PR's diff after the forward merge reaches `main`.

## Changes

- derive Linux and Windows producer namespaces from the committed `VERSION` file instead of `git describe`
- validate that `VERSION` uses `YY.MM.PP` format
- reconstruct all publishable Linux image tags from `matrix.yml`, including OS-qualified tags and OS-free aliases
- reject any checked-in RAPIDS unified devcontainer `BASE` that the current matrix cannot produce
- run that validation in pre-commit and as a fatal release prerequisite before Linux producers and downstream unified devcontainers run
- add ShellCheck coverage for the release-bootstrap scripts touched here

## Why the previous sequence cannot recur

After these changes, merging a version update changes both producer and consumer namespaces through the same committed `VERSION` value. The producer no longer depends on whether the alpha tag existed when the workflow checkout occurred. The release also stops before building images if a consumer base does not correspond to the current matrix, so component-version drift is reported directly rather than surfacing as a late missing-manifest error.

## Validation

- `pre-commit run --all-files`
  - zizmor
  - devcontainer base-image validation
  - ShellCheck for the release-bootstrap scripts
- positive validation against the 26.12 / UCX 1.21 configuration
- negative validation with a deliberate UCX 1.19 base, which failed and identified the affected file and unproducible tag
- `git diff --check`

Incident: https://github.com/rapidsai/devcontainers/actions/runs/34280170418/job/102279620120

### DIFF
--- a/.github/actions/devcontainer-json/action.sh
+++ b/.github/actions/devcontainer-json/action.sh
@@ -1,13 +1,17 @@
 #! /usr/bin/env bash
 
 # cd to the repo root
-cd "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/../../../";
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+cd "${script_dir}/../../../" || exit 1
 
 os="${1:-"ubuntu:22.04"}";
 features="${2:-"[]"}";
 container_env="${3:-"null"}";
 
-VERSION="$(git describe --abbrev=0 --tags --first-parent | sed 's/[a-zA-Z]//g' | cut -d '.' -f -2)";
+# VERSION is updated before the alpha tag is created during release rollover.
+# Use it as the source of truth so a release workflow triggered by that update
+# publishes the new image namespace instead of the previous tagged version.
+VERSION="$(cut -d '.' -f 1-2 VERSION)";
 tag="$(node -p "$(cat <<EOF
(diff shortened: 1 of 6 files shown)


bdice (approved):
I agree with all of these changes.

trxcllnt (approved):

PR #770, opened 2026-09-09
add pre-commit check for required version bumps

msarahan:
It is common for people to change code in a feature and forget to bump the version. When that happens, the new images are not pushed to GHCR. This check is intended to assist with that.

It accounts for the following possible situations:
* forgotten version bump
* version value lower than high watermark for XX.YY release series
* no-op if manual version bump has been done

This has one config setting - whether to automatically fix files, like a linter.

```
      # Opt in to automatic feature version updates in this checkout with:
      # git config devcontainers.auto-bump-feature-versions true
```

### DIFF
--- a/features/src/rapids-build-utils/devcontainer-feature.json
+++ b/features/src/rapids-build-utils/devcontainer-feature.json
@@ -1,7 +1,7 @@
 {
   "name": "NVIDIA RAPIDS devcontainer build utilities",
   "id": "rapids-build-utils",
-  "version": "26.12.0",
+  "version": "26.12.1",
   "description": "A feature to install the RAPIDS devcontainer build utilities",
   "containerEnv": {
     "BASH_ENV": "/etc/bash.bash_env"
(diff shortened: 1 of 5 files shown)


features/src/rapids-build-utils/devcontainer-feature.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "NVIDIA RAPIDS devcontainer build utilities",
   "id": "rapids-build-utils",
-  "version": "26.12.0",
+  "version": "26.12.1",
```
trxcllnt, line 4: Does this version need to be bumped in this PR?

msarahan: I don't understand why it was added. I'll look into why this happened.

msarahan: This is a catch-up bump from the forward-merge in #768 . It should have been done there. Would you like me to split it off into its own PR, since it is noise in this PR?

PR #778, opened 2026-09-11
Bump msvc build tools to 14.51

davebayer:


windows/image/installers/install-vs.ps1
```diff
@@ -26,7 +26,8 @@ $vsComponentVersion = @{
     "14.43"  = "14.43.17.13"
     "14.44"  = "14.44.17.14"
     "14.50"  = "14.50.18.0"
-    "latest" = "14.50.18.0"
+    "14.51"  = "14.51.18.6"
```
davebayer, line 29: I'm not entirely sure about the `14.51.18.6` value, I don't know where to get it from. I suppose it's the build tools version + the vs version and it was released in 18.6

PR #779, opened 2026-09-16
Switch RAPIDS devcontainers to Ubuntu 26.04

bdice:
Switch the shared RAPIDS image base from Ubuntu 24.04 to Ubuntu 26.04, covering CUDA and Mambaforge devcontainers. Leave CCCL images and feature-test baselines unchanged.


msarahan (approved):
I'm fine with this, but I don't use devcontainers frequently and don't understand the possible impact here. Please check CI carefully before merging.

PR #780, opened 2026-09-16
Use default dependency keys for cuML

bdice:
Use cuML’s default dependency keys instead of its custom `devcontainers` key now that conda-forge XGBoost no longer depends on RMM.

Merge and publish this feature before NVIDIA/cuml#8660.

Part of rapidsai/build-planning#312.


msarahan (approved):

bdice:
Failures look like they're in cuVS, something cutile-related. That shouldn't be a blocker for this change. Merging.

trxcllnt (approved):

KyleFromNVIDIA (approved):

jakirkham:
Thanks Bradley! 🙏

We are seeing [the following error in CI]( https://github.com/rapidsai/devcontainers/actions/runs/35117356454/job/104866128077?pr=780#step:9:2520 ):

```
Not creating 'rapids' conda environment because 'rapids.yml' is empty.
```

Is there something else we are missing from this change (or cuML)?

bdice:
That isn't the true error. That `'rapids.yml' is empty.` happens before repos are cloned and the environment is created. The failures are cuVS related, which I mentioned above. That is not a blocker for this to merge.

bdice:
I think https://github.com/NVIDIA/cuml/pull/8660 contains a fix for the cuTile issue, but it depended on this PR merging first.

